PR #64, opened 2023-07-06
fix(module): validate npm package before installing

arashsheyda:
tried to help with the TODO

also is it a good idea to check if the module is a Nuxt Module and if so, add it to nuxt.config?

https://github.com/nuxt/cli/blob/442fab890aa1419cb3830e81550c0fef336d989c/src/commands/module/add.ts#L46-L48


src/commands/module/add.ts
```diff
@@ -29,8 +30,12 @@ export default defineCommand({
   async setup(ctx) {
     const cwd = resolve(ctx.args.cwd || '.')
 
-    // TODO: Resolve and validate npm package name first
     const npmPackage = ctx.args.moduleName
+    const isValid = await validateNpmModule(npmPackage) 
+
+    if (!isValid) {
+      return consola.error(`\`${npmPackage}\` is not a valid npm package`)
```
pi0, line 37: We might use `throw new Error` instead for better control flow and proper exit status code.

arashsheyda: aw yes, thanks

src/commands/module/_utils.ts
```diff
@@ -71,3 +71,12 @@ export async function fetchModules(): Promise<NuxtModule[]> {
   )
   return data
 }
+
+export async function validateNpmModule(module: string): Promise<boolean> {
+  try {
+    await $fetch(`https://registry.npmjs.org/${module}`)
```
pi0, line 77: One thing i am worried is that this way we don't support private packages/registries. We might only use a soft warning to say hey we couldn't validate the package from public npm registry, it might go wrong.

pi0:
Hi. Sorry for being late back on this. Checking locally, since now we show pm output, they expect to show more useful information about why package could not be fetched (it could be network access or lack of permissions for private registries).

Also it is faster path to try installing by default and add fetch latency when we fail.

image

For now i think best is to rely on package manager's DX for displaying better errors. (and also rely on modules DB)

arashsheyda:
@pi0 hi, sorry I didn't see this earlier. Yes! I think this is much better. thanks

PR #66, opened 2023-07-07
feat: warn when node version is incompatible

### LINKED ISSUE #52: Automatic check and alert when using too old version of Node
### Describe the feature

I think it would make sense if Nuxt notifies the user if it is using a version of Node that is too old for nuxt3 during start up.
Just a cheaky little text explaining what version the user is on and what the recommended version is probably enough.

I just spent far too much time to admit trying to debug a weird issue with my project that turned out to be because I was on an old version of node 16. It works fine in 18. I have to constantly switch between when going between projects.

### Additional information

- [X] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://v3.nuxtjs.org/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/framework/discussions) and [issues](https://github.com/nuxt/framework/issues).

nexxeln:
Closes #52 

- added a utility function `isCompatibleNodeVersion`
- added checks to build and dev commands and warnings if the version is incompatible

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -7,6 +7,7 @@ import { overrideEnv } from '../utils/env'
 import { showVersions } from '../utils/banner'
 import { defineCommand } from 'citty'
 import { sharedArgs, legacyRootDirArgs } from './_shared'
+import { isCompatibleNodeVersion } from '../utils/nodeVersion'
 
 export default defineCommand({
   meta: {
@@ -27,6 +28,11 @@ export default defineCommand({
   },
   async run(ctx) {
     overrideEnv('production')
+    if (!isCompatibleNodeVersion()) {
+      consola.warn(
+        'You are using an outdated version of Node.js. Please upgrade to Node.js >= 16.10.0.',
+      )
+    }
 
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir || '.')
 
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -17,6 +17,7 @@ import { clearBuildDir } from '../utils/fs'
 import { defineCommand } from 'citty'
 
 import { sharedArgs, legacyRootDirArgs } from './_shared'
+import { isCompatibleNodeVersion } from '../utils/nodeVersion'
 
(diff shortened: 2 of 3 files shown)


src/commands/build.ts
```diff
@@ -27,6 +28,11 @@ export default defineCommand({
   },
   async run(ctx) {
     overrideEnv('production')
+    if (!isCompatibleNodeVersion()) {
+      consola.warn(
```
dreyfus92, line 32:
```suggestion
      console.warn(
```

src/commands/dev.ts
```diff
@@ -69,6 +70,12 @@ export default defineCommand({
   async run(ctx) {
     overrideEnv('development')
 
+    if (!isCompatibleNodeVersion()) {
+      consola.warn(
```
dreyfus92, line 74:
```suggestion
      console.warn(
```

dreyfus92 (changes requested):
I leave you a couple of small suggestions @nexxeln 😁

pi0:
Hi and thanks for this PR. The current Node.js engine version check is already implemented here:

https://github.com/nuxt/cli/blob/main/src/utils/engines.ts


We globally check the engines (Node.js version) in background for all comments via `setup` hook in `main.ts`:

https://github.com/nuxt/cli/blob/main/src/main.ts#L20

The engines field is defined in `package.json`:

https://github.com/nuxt/cli/blob/main/package.json#L83

So this is in place. As an improvement, we might support a custom range parameter to `checkEngines` and move this utility to `@nuxt/kit` as @danielroe suggested so that we can also check range in `nuxt/package.json`.

Feel free to rework on this PR or move it to nuxt/nuxt repository.

src/commands/build.ts
Intevel: @dreyfus92  consola is right, its the library of unjs.

nexxeln:
Sounds good, completely missed the implementation in engines.ts file. I'll move this PR to nuxt/nuxt

PR #88, opened 2023-08-24
feat: use listhen to parse its own `dev` args

danielroe:
type issue resolved by https://github.com/unjs/citty/pull/56

We might consider deprecating sslKey and sslCert as well?


pi0 (approved):
Thanks!

pi0:
> We might consider deprecating sslKey and sslCert as well?

I wish... It is just too late as people might be using it. But surely we can think about that to make nuxi simpler.

PR #111, opened 2023-08-28
refactor(upgrade): use `--force` by default

pi0:
Most of the times and by default we recommend `--force` with the upgrade command. I think it is sensible to have it enabled by default and enable `--no-force` for opt-out.

image

### DIFF
--- a/src/commands/upgrade.ts
+++ b/src/commands/upgrade.ts
@@ -35,6 +35,7 @@ export default defineCommand({
     force: {
       type: 'boolean',
       alias: 'f',
+      default: true,
       description: 'Force upgrade to recreate lockfile and node_modules',
     },
   },


danielroe:
In honesty, I don't think it's a good pattern to remove `node_modules` and fully recreate lockfile (although it has been necessary), as it has considerable side-effects for _non_ Nuxt dependencies. We are working around package manager bad behaviour but with latest pnpm release this may be less of an issue.

If we can possibly detect problems with nuxt/unjs deps I would hope we may be able to solve this in future with a less nuclear option.

Personally, I don't ever use `--force` but instead upgrade + then use `pnpm dedupe` or other package manager commands.

pi0:
AFAIK latest pnpm v8 only reverted behavior for top level ranges to be same as v7 and respect latest.

Indeed it is not the best idea to recreate specially in monorepos.

But also not recreating makes much more hoisting issues up to a level we everywhere are recommending to use —force

Idea behind this change is that if we are adding force in the upgrade guides (and might remove it in the future) let the nuxi change its behavior once pnpm was ready or once we found hosing issue solution.

wdyt that when unspecified (neither true or false, have a prompt to ask this?

src/commands/upgrade.ts
```diff
@@ -57,9 +58,23 @@ export default defineCommand({
     consola.info('Current nuxt version:', currentVersion)
 
     // Force install
+    const pmLockFile = resolve(cwd, packageManagerLocks[packageManager])
+    const forceRemovals = ['node_modules', relative(process.cwd(), pmLockFile)]
+      .map((p) => colors.cyan(p))
+      .join(' and ')
+    if (ctx.args.force === undefined) {
+      ctx.args.force = await consola.prompt(
+        `Would you like to recreate ${forceRemovals} to fix problems with hoised dependency versions are ensure you have the most up-to-date dependencies?`,
```
danielroe, line 67:
```suggestion
        `Would you like to recreate ${forceRemovals} to fix problems with hoisted dependency versions and ensure you have the most up-to-date dependencies?`,
```

src/commands/upgrade.ts
```diff
@@ -57,9 +58,23 @@ export default defineCommand({
...
+          type: 'confirm',
+          default: true,
+        },
+      )
+    }
     if (ctx.args.force) {
-      consola.info('Removing lock-file and node_modules...')
-      const pmLockFile = resolve(cwd, packageManagerLocks[packageManager])
+      consola.info(
+        `Recreating ${forceRemovals}. If you have encountred any issues, revert the changes and try with \`--no-force\``,
```
danielroe, line 76:
```suggestion
        `Recreating ${forceRemovals}. If you encounter any issues, revert the changes and try with \`--no-force\``,
```

danielroe (approved):

PR #119, opened 2023-08-29
fix(devtools): define `enable|disable` command as positional argument

### LINKED ISSUE #69: `nuxt devtools enable|disable` works only with `--command` flag
## Problem
According to the [docs](https://nuxt.com/docs/api/commands/devtools), users can just type `nuxi devtools enable` to enable devtools in the project. Howerver, after https://github.com/nuxt/cli/pull/47 enable/disable flag can only be submitted with `--command` before the value.

There I need to ask: which way is preferred for the future? Should we change documentation or fix the command itself?

### 2. There I logged `ctx.args`:
image

### 3. There is the run without `--command` flag, as documented:
image

### 4. There is the run with `--command flag`:
image

AndreyYolkin:
Resolves #69 

This PR registers `command` as positional argument and reorders it before legacy `[rootDir]` option to align behavior with described in [docs](https://nuxt.com/docs/api/commands/devtools)

Also it changes defined command meta to give right output when user types `nuxi devtools` without specifying the action.

### DIFF
--- a/src/commands/devtools.ts
+++ b/src/commands/devtools.ts
@@ -11,12 +11,12 @@ export default defineCommand({
   },
   args: {
     ...sharedArgs,
-    ...legacyRootDirArgs,
     command: {
-      type: 'string',
+      type: 'positional',
       description: 'Command to run',
       valueHint: 'enable|disable',
     },
+    ...legacyRootDirArgs,
   },
   async run(ctx) {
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks!

danielroe (approved):

pi0:
Released in [v3.7.3](https://github.com/nuxt/cli/releases/tag/v3.7.3)

PR #120, opened 2023-08-29
fix(dev): pass listener interface to the `listen` hook

### LINKED ISSUE #117: [Hooks] `listen` does not send the `listener` as 2nd arg
Since Nuxt 3.7.0

Found in https://github.com/nuxt-modules/tailwindcss/blob/812ce3981458e3f1cd2b9b823c68ecb5ada07ef8/src/viewer.ts#L23

pi0:
resolves #117

Since we now listen on the main process, this PR makes modules that depending on `listen` hook keep working.

**Note:** This is not a stable solution. Modules should not depend deeply on internal server listeners. 

Also, we pass the server without HTTPs to the vite-node server to avoid certificate issues as we had before.

--- 

Verified on tailwind module:

image

### DIFF
--- a/src/commands/dev-internal.ts
+++ b/src/commands/dev-internal.ts
@@ -15,6 +15,7 @@ import { RequestListener, Server } from 'http'
 import { toNodeListener } from 'h3'
 import { AddressInfo } from 'net'
 import { isTest } from 'std-env'
+import { Listener } from 'listhen'
 
 export type NuxtDevIPCMessage =
   | { type: 'nuxt:internal:dev:ready'; port: number }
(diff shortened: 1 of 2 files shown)


pi0:
Merging to move forward quickly and introduce hotfix. Internally we need to revise `listen` hook usage behavior in the next steps.

pi0:
Merging to move forward quickly and introduce hotfix. Internally we need to revise `listen` hook usage behavior in the next steps.

Atinux:
Happy to release as `3.7.3`?

pi0:
Released in [v3.7.3](https://github.com/nuxt/cli/releases/tag/v3.7.3)

PR #123, opened 2023-08-30
fix: add cli version and name in usage

pi0:
We were missing nuxi version and description in usage

image


pi0:
(i think we can refine description later, PR and ideas welcome!)

PR #124, opened 2023-08-30
fix: improve preview banner and add backward compatible `start`

### LINKED ISSUE #113: `nuxi start` missing as a part of command cli in v3.7
In the previous versions of nuxt, I was using `nuxt start` to start my production server, though I did later find out it's just an alias to `nuxt preview` after looking at the source code. It is convenient to have this alias though, could we add this back?

pi0:
Resolves #113 by adding backward compatible `start` command.

Also improves the Preview mode experience with explicit warnings about production usage.

image


Atinux (approved):

Atinux:
It. Is. Beautiful 😍

Issue #108, opened 2023-08-28
WebSocket Support 


pi0:
[description is wip]

Since Nuxt 3, we had (semi-official) solutions to support WebSocket in development and production but never a standard layer. This had been something we planned to support universally in nuxt/nitro dev server and runtime.

With the introduction of new dev server CLI architecture in nuxi 3.7 (#81) some of the current solutions stooped to be working.

My plan is to first make sure there is a basic replacement for current solutions and then put the stable WS layer on fast track to finally resolve this. (Update: hooking to the server `upgrade` via `listen` should work again now via https://github.com/nuxt/cli/pull/127)

Related issues:
- #101
- #107
- #112

mumi:
Any update on this?

BghinC:
> Any update on this?

We just got an update yesterday. You can learn more about it here: https://twitter.com/_pi0_/status/1761875980849713261

pi0:
Experimental websocket is available in nightly channels. 

See https://github.com/pi0/nuxt-websocket

pi0:
CLI support landing: https://github.com/nuxt/cli/pull/348

Note: Some features such as supporting `devHandlers` are not complete yet and need progress on Nitro side. You can track via https://nitro.unjs.io/guide/websocket and https://github.com/unjs/nitro/issues/2171

PR #127, opened 2023-08-30
fix(dev): support ws upgrades

pi0:
#108

#101

This PR adds support for `upgrade` event from dev server wrapper, proxying it to the underlying (build time) userland server instance.

I have also added a simple playground page to try things out (it is obviously not a right usage to register ws but current workaround)

image


Atinux:
🚀

PR #134, opened 2023-08-31
feat(mod-search): filter mod search by compatible nuxt version

### LINKED ISSUE #115: `nuxi module search` should filter incompatible module with current Nuxt version

rohrig:
Closes #115 

❓ Type of change
 ✨ New feature (a non-breaking change that adds functionality)

📚 Description
This PR introduces a version compatibility filter for the module search function.


src/commands/module/search.ts
```diff
@@ -18,15 +20,21 @@ export default defineCommand({
       description: 'keywords to search for',
       required: true,
     },
+    nuxtVersion: {
+      type: 'string',
+      description: 'Filter by Nuxt Version and only see compatible modules',
```
Atinux, line 25:
```suggestion
      description: 'Filter by Nuxt version and list compatible modules',
```

Atinux:
Thank you!

Wondering if we should leverage https://github.com/nuxt/nuxt/blob/5390ca43212656b7f169153024f9b19bcd10ec25/packages/kit/src/compatibility.ts#L12 or not 🤔 

It would be nice to automatically detect the current Nuxt version installed in the project to give as as default if `ctx.args.nuxtVersion` is not defined, wdyt @rohrig ?

rohrig:
> Thank you!
> 
> Wondering if we should leverage https://github.com/nuxt/nuxt/blob/5390ca43212656b7f169153024f9b19bcd10ec25/packages/kit/src/compatibility.ts#L12 or not 🤔
> 
> It would be nice to automatically detect the current Nuxt version installed in the project to give as as default if `ctx.args.nuxtVersion` is not defined, wdyt @rohrig ?

Hi @Atinux ,

I think that's ideal. It was my initial goal but I wasn't able to figure it out and got distracted 😄 

I'd be happy to implement it if you give me some pointers on how to do it. 

I attempted to do this but I ran into the error that Nuxt is not available:
image

image

pi0 (approved):
Thanks! I have made few refactors + auto detected nuxt version

PR #153, opened 2023-09-07
refactor(dev): rewrite to support `--no-fork` and improve stability

pi0:
This PR applies relatively large amounts of refactoring to the dev command:

- Extract base nuxt dev server logic to shared util
- Support `--no-fork` mode without spawning a new process
- Improved loading handler by eagerly loading template and providing nitro dev handler all the times 

There might be some small regressions with this changes but i think overall better stability is worth to try and iterate faster. Checking locally little bit more to make sure there are not odd issues.

TODO: I am working to always initialize an internal listener + proxy. This patterns helps on HTTPs + Expose host stability for vite-node and internal calls.


playground/pages/ws.vue
```diff
@@ -2,7 +2,11 @@
 import { WebSocket } from 'unws'
 
 const reqURL = useRequestURL()
-const urls = [`ws://${reqURL.host}/api/ws`, 'ws://localhost:8080']
+const isSecure = reqURL.protocol === 'https:'
```
pi0, line 5: TODO: This casues hydration mismatch because is secure util is not working properly. Unrelated issue to solve later

PR #184, opened 2023-09-14
feat(dev): reuse main dev port for vite hmr

pi0:
This PR enables Nuxt to use main dev server as HMR server instead of requiring an external port.

Instead of updating Nuxt, using hook in nuxi allows moving forward faster and keep it closer to the listener.

~~Current issues: Vite keeps hooking on internal web server for ws. After a (soft) reload, hmr will be broken~~

Currently, seems quite fast and stable. 

https://github.com/nuxt/cli/assets/5158436/b23573f9-61d1-4611-bac6-73cc511f69dd

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -153,6 +153,13 @@ class NuxtDevServer extends EventEmitter {
       (config, { isClient }) => {
         if (isClient && config.server) {
           config.server.hmr = {
+            ...(config.server.hmr as Exclude<
+              typeof config.server.hmr,
+              boolean
+            >),
+            protocol: undefined,
+            port: undefined,
+            host: undefined,
             server: this.listener.server,
           }
         }
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -147,6 +147,18 @@ class NuxtDevServer extends EventEmitter {
       },
     })
 
+    // Connect Vite HMR
+    this._currentNuxt.hooks.hookOnce(
+      'vite:extendConfig',
+      (config, { isClient }) => {
+        if (isClient && config.server) {
+          config.server.hmr = {
+            server: this.listener.server,
+          }
+        }
+      },
+    )
+
     // Write manifest and also check if we need cache invalidation
     if (!reload) {
       const previousManifest = await loadNuxtManifest(
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -158,6 +158,9 @@ class NuxtDevServer extends EventEmitter {
         }
       },
     )
+    this._currentNuxt.hooks.hookOnce('close', () => {
+      this.listener.server.removeAllListeners('upgrade')
+    })
 
     // Write manifest and also check if we need cache invalidation
     if (!reload) {


antfu:
That's pretty cool :)

pi0:
Yep, it should partially be useful for remote scenarios such as codespaces, wsl and local network

src/utils/dev.ts
```diff
@@ -147,6 +147,21 @@ class NuxtDevServer extends EventEmitter {
...
     })
 
+    // Connect Vite HMR
+    this._currentNuxt.hooks.hookOnce(
+      'vite:extendConfig',
+      (config, { isClient }) => {
+        if (isClient && config.server) {
+          config.server.hmr = {
+            server: this.listener.server,
+          }
```
danielroe, line 157: is it worth merging configs in case user has set `overlay` or `timeout` or otherwise configured this?

pi0: Mainly i wanted to avoid vite confliciting with (nuxt provided) port options. Let me try.

danielroe: We can update nuxt to disallow setting certain options in `vite.server.hmr` if we want to take full control of them. of course in most cases users will not configure this but there may be some edge cases

pi0: we only nullify intentional keys now: cad706b6696f1c335da266e57d65f91dba1d4d5a

Surely would be nice to also limit in types if we finally want to support main port 👍🏼

pi0:
Let's try on edge

martinsjek:
@pi0 

Can you give me an example what to change when using Apache/Nginx while developing?

Before 3.9.0 version everything worked fine for at least a year :D

Here is my Apache:
`
<IfModule mod_ssl.c>
    <VirtualHost *:443>
         ServerAlias my.local.io

         SSLEngine on
         SSLCertificateFile      /etc/apache2/ssl/my.crt
         SSLCertificateKeyFile   /etc/apache2/ssl/my.key

        ProxyPass /_nuxt/hmr/ ws://docker-container:3001/
        ProxyPassReverse /_nuxt/hmr/ ws://docker-container:3001/
        RewriteEngine on
        RewriteCond %{HTTP:Upgrade} websocket [NC]
        RewriteCond %{HTTP:Connection} upgrade [NC]
        RewriteRule ^/?(.*) "ws://docker-container:3001/$1" [P,L]

        ProxyPass / http://docker-container:3000/
        ProxyPassReverse / http://docker-container:3000/
    </VirtualHost>
</IfModule>
`

Here is my nuxt config:
`
...
vite: {
    server: {
      hmr: {
        protocol: 'wss',
        port: 3001,
        clientPort: 443,
        path: 'hmr/',
      },
    },
  },
...
`

wanshot:
@martinsjek 

Have you tried setting the NUXI_DISABLE_VITE_HMR environment variable? This might allow you to continue using your setup without changes, as you did before the update to version 3.9.0

ref: https://github.com/nuxt/cli/pull/189

martinsjek:
Thanks. I have already resolved this issue with the help of **pi0** .

PR #199, opened 2023-09-19
fix: soft warning for `node < 18`

pi0:
- ~~Remove 14.x from engines filed because we actually can't support it anymore~~ (delayed until nuxt does it too!)
- Uses a soft warning for 16.x since Node 16 is EOL
- Remove `_dev` command from checks to avoid duplicate warn and `stackblitz` runtime since users have no control!

image


package.json
```diff
@@ -79,6 +79,6 @@
   },
   "packageManager": "pnpm@8.7.6",
   "engines": {
-    "node": "^14.18.0 || >=16.10.0"
+    "node": ">=16.10.0"
```
danielroe, line 82: Is this change strictly required?

src/utils/engines.ts
```diff
@@ -1,16 +1,14 @@
-import nuxiPkg from '../../package.json'
-
 export async function checkEngines() {
   const satisfies = await import('semver/functions/satisfies.js').then(
     (r) =>
       r.default || (r as any as typeof import('semver/functions/satisfies.js')),
   ) // npm/node-semver#381
   const currentNode = process.versions.node
-  const nodeRange = nuxiPkg.engines.node
+  const nodeRange = '>= 18.0.0'
```
danielroe, line 7: could we respect the `engines.node` in the nuxt package, perhaps, to allow for changes here without a nuxi upgrade?

pi0: Constraints of CLI are not exactly the same as Nuxt Core We might need a version constraint that is only needed for CLI core. 

This also requires the overhead of resolving all possible nuxt packages before every command.

package.json
pi0: Nuxi CLI bundle (today) is compatible but it is so likely we will break it unnoticed with any of the bundled deps.

Also nuxt/nitro and even pnpm are incompatible with any command! It just makes sense to let users be aware of this as a strict requirement: (and there is always escape hack with skip-engines flag for rare conditions someone want to make Nuxt 3 compatible with Node 14...)

image

danielroe (approved):

PR #227, opened 2023-09-25
chore: add script to run cli in playground context

warflash:
Trying to get familiar with the CLI at the moment and couldn't find an easy way to use the actual local CLI build for the playground.
Very well possible I just missed the obvious and existing way to go about that process 😅

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "dev:bun": "bun --bun ./bin/nuxi.mjs dev ./playground",
     "lint": "eslint . && prettier --check src",
     "lint:fix": "eslint --fix . && prettier --write src",
+    "play:nuxi": "cd ./playground && node ../bin/nuxi.mjs",
     "nuxi": "node ./bin/nuxi.mjs",
     "nuxi-bun": "bun --bun ./bin/nuxi.mjs",
     "prepack": "unbuild",


package.json
```diff
@@ -27,6 +27,7 @@
     "dev:bun": "bun --bun ./bin/nuxi.mjs dev ./playground",
     "lint": "eslint . && prettier --check src",
     "lint:fix": "eslint --fix . && prettier --write src",
+    "play:nuxi": "cd ./playground && node ../bin/nuxi.mjs",
```
pi0, line 30:
```suggestion
    "play": "./bin/nuxi.mjs --cwd ./playground",
```

warflash: Thanks!
Played with that a bit and it seems like we need to pass commands before the --cwd flag, so e.g.
`node ./bin/nuxi.mjs analyze --cwd ./playground`, but then we loose the ability to feed the command we want at the console level.
With the original suggestion you were able to execute `pnpm play cleanup` or any other nuxi command directly.
Hope that makes sense

pi0: Yeah thanks for checking. I think there is area to improve for this to make it possible.

rohrig: @warflash 
I had the same struggle when first playing with the cli locally. Turns out the simplest solution I could find was to run it like so:

from the playground dir:

```bash
    ../bin/nuxi.mjs <whatever normally come after running nuxi>
```

pi0: Yes, that's what i also normally do today. I think point of this PR, (which is also valid for real projects) is to prepare a generic command that allows nuxi operating in separate directory. There are few different ways such as `cd` to the dir or directly `nuxi <dev> [dir]` (limitation is, some command only accept `--dir`)

pi0:
Tracking via https://github.com/nuxt/cli/issues/365

PR #230, opened 2023-09-26
fix: call `nuxt.ready()` before registering hooks

chz:
@pi0 please check. Fixed issue: https://github.com/nuxt/bridge/issues/918

this happen only 3.9.0. But 3.8.4 work normally

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -207,7 +207,9 @@ class NuxtDevServer extends EventEmitter {
     }
 
     // Remove websocket handlers on close
-    this._currentNuxt.hooks.hook('close', () => {
+    // nuxt bridge doesn't have support for `hookOnce` here
+    const unsubClose = this._currentNuxt.hooks.hook('close', () => {
+      unsubClose()
       this.listener.server.removeAllListeners('upgrade')
     })
 
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -184,7 +184,7 @@ class NuxtDevServer extends EventEmitter {
         ...this.options.overrides,
       },
     })
-
+    await this._currentNuxt.ready();
     // Connect Vite HMR
     if (!process.env.NUXI_DISABLE_VITE_HMR) {
       this._currentNuxt.hooks.hook(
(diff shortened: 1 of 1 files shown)


src/utils/dev.ts
```diff
@@ -207,7 +207,7 @@ class NuxtDevServer extends EventEmitter {
     }
 
     // Remove websocket handlers on close
-    this._currentNuxt.hooks.hookOnce('close', () => {
+    this._currentNuxt.hooks.hook('close', () => {
       this.listener.server.removeAllListeners('upgrade')
```
danielroe, line 211:
```suggestion
    const unsubClose = this._currentNuxt.hooks.hook('close', () => {
      unsubClose()
      this.listener.server.removeAllListeners('upgrade')
```

pi0: Is there a chance we might polyfill hookOnce from once place? Ideally in core or bridge but since i support requires more work, in kit loader at least to patch and support `hookOnce`. Making sure bridge keeps working in long term via this patches seems less practical to me.

danielroe: That would be better, for sure.

pi0: Separate topic how long we should support nuxt 2 / bridge for nuxi we can chat later -- specially as first class (which I'm not sure is also being tested against nuxt/nuxt/kit)

In the meantime yeah let's find a better way to make sure at least regressions with hookOnce won't happen in a safer way.

danielroe: we do test bridge against nuxt with ecosystem-ci before every release of nuxt

src/utils/dev.ts
```diff
@@ -207,7 +207,9 @@ class NuxtDevServer extends EventEmitter {
     }
 
     // Remove websocket handlers on close
-    this._currentNuxt.hooks.hookOnce('close', () => {
+    // nuxt bridge doesn't have support for `hookOnce` here
+    const unsubClose = this._currentNuxt.hooks.hook('close', () => {
+      unsubClose()
```
danielroe, line 212: With https://github.com/nuxt/bridge/pull/921, I think we can revert this change and just keep the reordered `ready()` call.

```suggestion
    this._currentNuxt.hooks.hookOnce('close', () => {
```

src/utils/dev.ts
```diff
@@ -184,7 +184,7 @@ class NuxtDevServer extends EventEmitter {
         ...this.options.overrides,
       },
     })
-
+    await this._currentNuxt.ready()
```
danielroe, line 187:
```suggestion
    await this._currentNuxt.ready()

```

PR #248, opened 2023-10-12
fix: support `-nightly` nuxt releases and publish `nuxi-nightly`

danielroe:
part of https://github.com/nuxt/nuxt/issues/22143

needs to be merged before: https://github.com/nuxt/nuxt/pull/23508

related: https://github.com/nuxt/cli/commit/d1120b1


pi0 (approved):
nice

PR #250, opened 2023-10-17
feat: detecting if cwd contains a nuxt project

### LINKED ISSUE #203: Detecting if cwd doesn't contain a Nuxt project
Default behavior for using the CLI in a directory without `nuxt.config.ts` is installing, then creating the Nuxt config file and adding the module. Can we check if `nuxt.config.ts` exists first ? 

Thinking behavior should be either showing a YES/NO question to create it and proceed? Or, just killing the process.

adhamfarrag:
This PR is dedicated to checking weather `nuxt.config.ts` exists or not. If not it fires a consola prompt to ask users if they want to proceed or not. Screenshot below for a demo. 

2023-10-17 at 6  59 10@2x

Fixes #203

### DIFF
--- a/src/commands/module/add.ts
+++ b/src/commands/module/add.ts
@@ -3,6 +3,7 @@ import { defineCommand } from 'citty'
 import { sharedArgs } from '../_shared'
 import { existsSync } from 'node:fs'
 import { loadFile, writeFile, parseModule, ProxifiedModule } from 'magicast'
+import { exists } from '../../utils/fs'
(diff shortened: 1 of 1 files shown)


adhamfarrag:
@pi0, this is not related to the scope of PR, but I noticed when it comes to installing a package that doesn't exist or private one with no permissions. There are several types of errors coming from nypm that some of them can be silenced or handled via Nuxi if nypm allows it. What do you think ?

PR #258, opened 2023-10-21
feat: add commands to analyse payloads

### LINKED ISSUE #257: Add a way to debug `_payload.json`
Hello,

In Nuxt3, all `useFetch` or `useAsyncData` push content in a `_payload` file when page is prerendered. This is cool but it's pretty easy to have some heavy payloads because of a mistake like keeping too many keys with Content.

Could be nice to introduce a new command to easily view every payload size.

Barbapapazes:
fix #257

Hello,

This is a first draft to fix #257. I'm not sure if it's the best way to do it.

In fact, we could have more options with this features

- add a warning and a error level in order to alert user if a file is too heavy (could throw an error in case of a file too heavy)
- add more formatting options
- add a way to only list ?
- add a way to sort by size or by file path

And maybe more options.

### DIFF
--- a/src/commands/payloads.ts
+++ b/src/commands/payloads.ts
@@ -1,0 +1,72 @@
+import { resolve } from "pathe";
+import { defineCommand } from "citty";
+import fs from "node:fs"
+
+import { legacyRootDirArgs, sharedArgs } from './_shared'
+
+export default defineCommand({
+  meta: {
+    name: "payloads",
+    description: "Used to analyse size of generated payloads"
+  },
+  args: {
+    ...sharedArgs,
+    ...legacyRootDirArgs,
+  },
+  async run(ctx) {
+    const cwd = resolve(ctx.args.cwd || ctx.args.rootDir || ".");
+
+    const outputDir = resolve(cwd, ".output");
+    const publicDir = resolve(outputDir, "public");
+
+    // List payloads files in .output/public (can be in subfolders so we call the function recursively)
+    const listPayloads = (sourceDir: string): string[] => {
+      const payloads: string[] = [];
+      const files = fs.readdirSync(sourceDir);
+      files.forEach(file => {
+        const filePath = resolve(sourceDir, file);
+
+        const isFile = fs.statSync(filePath).isFile();
+        if (isFile && file === "_payload.json") {
+          payloads.push(filePath);
+          return
+        }
+
+        const isDirectory = fs.statSync(filePath).isDirectory();
+        if (isDirectory) {
+          const subPayloads = listPayloads(filePath);
+          payloads.push(...subPayloads);
+          return
+        }
+      });
+      return payloads;
+    }
+
+    const payloads = listPayloads(publicDir);
+
+    // Get size of each payload
+    const payloadsSizes = payloads.map(payload => {
+      const payloadSize = fs.statSync(payload).size;
+      return {
+        payload,
+        payloadSize
+      }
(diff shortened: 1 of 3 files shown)


src/commands/payloads.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    // Display sorted payloads
+    sortedPayloads.forEach(payload => {
+      console.log(`${payload.payload} ${toHumainSize(payload.payloadSize)}`);
+    });
+    // TODO: improve output. We could use a table like or a tree like output (waiting for https://github.com/unjs/consola/pull/223).
+  },
+})
+
+function toHumainSize(bytes: number): string {
```
Hebilicious, line 67:
```suggestion
function toHumanSize(bytes: number): string {
```

src/commands/payloads.ts
```diff
@@ -0,0 +1,72 @@
...
+        payloadSize
+      }
+    });
+
+    // Sort payloads by size
+    const sortedPayloads = payloadsSizes.sort((a, b) => b.payloadSize - a.payloadSize);
+
+    // Display sorted payloads
+    sortedPayloads.forEach(payload => {
+      console.log(`${payload.payload} ${toHumainSize(payload.payloadSize)}`);
```
Barbapapazes, line 61:
```suggestion
      console.log(`${payload.payload} ${toHumanSize(payload.payloadSize)}`);
```

pi0:
Hi dear @Barbapapazes and thanks for making this PR. I think it would be a great idea to support payload analyzer but we could support it from different places and i guess we might need to iterate over idea.

What do you think to instead create `nuxt-payload-analyzer` package exposing `nuxt-analyze-payload` bin command (`nuxt analyze-payload` will work) and also possibly a nuxt module that exposes same ability for dev tools.

Once solution is proven, we might think to either include it to nuxt or nuxi core or add as an ad-hoc module.

Barbapapazes:
> Hi dear @Barbapapazes and thanks for making this PR. I think it would be a great idea to support payload analyzer but we could support it from different places and i guess we might need to iterate over idea.
> 
> What do you think to instead create `nuxt-payload-analyzer` package exposing `nuxt-analyze-payload` bin command (`nuxt analyze-payload` will work) and also possibly a nuxt module that exposes same ability for dev tools.
> 
> Once solution is proven, we might think to either include it to nuxt or nuxi core or add as an ad-hoc module.

Yes, I totally agree with the "need  to iterate over idea" because this idea is an early stage idea! 🙂 I'm ok for making an external package to try it! Should I create it in my own repo?

Barbapapazes:
Yes, I think we can close it and I will link the package as soon as it is created. (PR can be closed too)

Barbapapazes:
🙂

image

_will publish a first version soon_

Barbapapazes:
For example with my website: 

image

Barbapapazes:
repo: https://github.com/barbapapazes/nuxt-payload-analyzer
package: https://www.npmjs.com/package/nuxt-payload-analyzer

pi0:
Nice! Will check it in the morning love to see how you did it! (also sharing in internal discord)

Barbapapazes:
And because it's fun with a huge website, I've try with nuxt.com

image

Barbapapazes:
We could have a Nuxt module to have analyze after a build!

image

Barbapapazes:
It can even fail if a payload is too large
image

PR #264, opened 2023-10-27
fix(typecheck): use resolve `vue-tsc` and `typescript`  path

dargmuesli:
This PR is motivated by [the removal of shameless hoisting in Nuxt](https://github.com/nuxt/nuxt/issues/14146) :tada: 

It replaces the custom `tryResolveModule` utility with the export from `@nuxt/kit`. I hope it's synonymous :pray: 

Also, more importantly, it reworks the resolution of `vue-tsc`. Background being: imagine a layer defining `vue-tsc` and `typescript` as `dependency` (not `devDependency`!) so that projects using that layer do not have to import `vue-tsc` themself but can lint right away.

- First question: is that even meant to be supported :joy:  or how should layer dependencies be categorized (normal, dev, peer...)?

To achieve that - I tested it with one of my projects - I was able to make `vue-tsc` discoverable by calling the function **without `cwd` and with `/bin/vue-tsc.js` suffix** instead of `/package.json` :warning: 

- Second question: what am I bricking with this? Feels like that change could certainly lead to issues for other use cases, @pi0 you'll for sure be able to know and see opportunities for improvement :bulb: 

I hope this PR is a helpful contribution. If I got something wrong, just tell me, no worries, I'm always willing to learn! :heart:


src/utils/kit.ts
```diff
@@ -1,4 +1,6 @@
-import { importModule, tryResolveModule } from './esm'
+import { tryResolveModule } from '@nuxt/kit'
```
pi0, line 1: We use `tryResolveModule` to resolve kit 😅 I prefer to keep a standalone version of resolver in cli to avoid behavior conflicting.

pi0 (approved):

pi0:
Thanks!

PR #268, opened 2023-10-30
refactor: use magicast `.includes` helper

angelhdzmultimedia:
This pr replaces the for loop to check if the modules array contains the added module, with a more performant `Array.includes`.

```
"Array.includes time:" 30 "ms"
"Manual loop time:" 326 "ms"
"Array.includes time:" 27 "ms"
"Manual loop time:" 242 "ms"
```

If the purpose of the for loop is backward compatibility, ignore this pr. 

🔥💚

### DIFF
--- a/src/commands/module/add.ts
+++ b/src/commands/module/add.ts
@@ -72,11 +72,10 @@ export default defineCommand({
         if (!config.modules) {
           config.modules = []
         }
-        for (let i = 0; i < config.modules.length; i++) {
-          if (config.modules[i] === r.pkgName) {
-            consola.info(`\`${r.pkgName}\` is already in the \`modules\``)
-            return
-          }
+       
+        if (config.modules.includes(r.pkgName) {
+          consola.info(`\`${r.pkgName}\` is already in the \`modules\``)
+          return
         }
         consola.info(`Adding \`${r.pkgName}\` to the \`modules\``)
         config.modules.push(r.pkgName)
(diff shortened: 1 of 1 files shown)


src/commands/module/add.ts
```diff
@@ -72,11 +72,10 @@ export default defineCommand({
         if (!config.modules) {
           config.modules = []
         }
-        for (let i = 0; i < config.modules.length; i++) {
-          if (config.modules[i] === r.pkgName) {
-            consola.info(`\`${r.pkgName}\` is already in the \`modules\``)
-            return
-          }
+       
+        if (config.modules.includes(r.pkgName)) {
```
pi0, line 76: FYI, `config` is not a real JS array but an instance of MagicAST proxy. Calling `.includes`, runs [this method](https://github.com/unjs/magicast/blob/main/src/proxy/array.ts#L52) which almost has same performance.

But this is a nice refactor anyway since this feature was not initially available in magicast 👍🏼

angelhdzmultimedia: Thank you! Very kind. 💚🔥

 I was just checking the code to see how to contribute to allow multiple modules addition, when I saw that part and said "hey, I can make my first non-typo contribution! This is easy to fix!". GPTChat helped me with the benchmarks code 😂. Going to play with magicast, I'm using pathe in my projects.
 
 As always, great job you all have been doing with Nuxt/unjs.

PR #272, opened 2023-11-01
fix(init): use absolute path for installing dependencies

### LINKED ISSUE #213: nuxi init . error
`nuxi init ` command in current directory causes this error `ERROR  EROFS: read-only file system, open '/.npmrc'  `
the case doesn't happen when using a directory.

jits:
Fixes #213

---

I've tested this out locally using:

```shell
node {path_to_locally_built_cli_with_this_change}/bin/nuxi.mjs init . --force
```

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -110,9 +110,12 @@ export default defineCommand({
         })
 
     // Get relative project path
-    const relativeProjectPath = relative(process.cwd(), template.dir)
+    let relativeProjectPath = relative(process.cwd(), template.dir)
+    if (relativeProjectPath === '') {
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks for investigating root cause ❤️ I have made slightly different changes to prefer using absolute path everywhere other than next step message which should be safer for similar issues.

pi0:
The fix should be testable via `npx nuxi-nightly@latest init . --force`

jits:
Thanks @pi0 — your changes make sense.

I'll take a look at the upstream issue and see if I can contribute a fix

PR #277, opened 2023-11-03
fix(analyze): set path for client analyze results

pi0:
Working on #276 noticed that nuxt by default does not sets client dist path causing `stats.html` to be generated in rootDir/workspaceDir ([src](https://github.com/nuxt/nuxt/blob/main/packages/vite/src/plugins/analyze.ts#L29))


danielroe:
it should in the next patch: https://github.com/nuxt/nuxt/blob/cd76c617fa7fb081af50fda71473fcbd946698db/packages/schema/src/config/build.ts#L111-L121.

danielroe (approved):

PR #278, opened 2023-11-03
feat(dev): return listener url

mandarini:
When calling `await runCommand("dev");` programmatically, I would like to be able to get the `baseUrl`, the listening URL.

This will also require a change in `citty`, to return the result, here: https://github.com/unjs/citty/blob/cb27bed49371682904d85421439e31b8b3b17d24/src/command.ts#L64 specifically, to return the result of the `run` function.

```
    if (typeof cmd.run === "function") {
      return await cmd.run(context);
    }
```


mandarini:
@pi0 I submitted this PR, it is needed in Nx, to get the baseUrl, and expose it to cypress. It's not urgent or blocking for us, but I thought it would be a nice to have! :)

pi0:
Thanks for the citty PR that's a good one! For receiving listener URL(s), you might also use a `listen` nuxt hook it gives the listener context.

src/commands/dev.ts
```diff
@@ -69,6 +69,7 @@ const command = defineCommand({
       // Fork nuxt dev process
       const devProxy = await _createDevProxy(nuxtOptions, listenOptions)
       await _startSubprocess(devProxy, ctx.rawArgs)
+      return devProxy?.listener?.url
```
pi0, line 72: Same as citty, we might return `{ listener }` here too (also notice when using a public URL or domain or tunnel, there are multiple listener URLs as well. `url` is the _prefered_ one we usually show first and used for public/qr when available)

mandarini: @pi0 I pushed a new commit with the change you requested! Thanks!

pi0 (approved):

pi0:
Thanks!

PR #290, opened 2023-11-15
feat(build): support `--preset`

### LINKED ISSUE #289: Support `--preset` for build command
Support `nuxi build --preset ...` as alternative to `SERVER_PRESET=... nuxi build` (nitro cli support `nitro build --preset` too)

pi0:
resolves #289

Support `npx nuxi build --preset <name>`

### DIFF
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -19,6 +19,10 @@ export default defineCommand({
       type: 'boolean',
       description: 'Build nuxt and prerender static routes',
     },
+    preset: {
+      type: 'string',
+      description: 'Nitro server preset',
+    },
     dotenv: {
       type: 'string',
       description: 'Path to .env file',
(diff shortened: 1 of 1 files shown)


src/commands/build.ts
```diff
@@ -44,7 +54,9 @@ export default defineCommand({
         logLevel: ctx.args.logLevel as 'silent' | 'info' | 'verbose',
         // TODO: remove in 3.8
```
Atinux, line 55: Shall we remove it now we have 3.8?

danielroe (approved):

Atinux (approved):

src/commands/build.ts
pi0: CLI is backward compatible i guess it would be harmless to keep and remove comment to mention why we have it extra here.

PR #300, opened 2023-11-24
fix: handle spaces in path at repo initialization

Ettapp:
This PR fix the issue happening when one run `nuxi init .` in a directory whose path contains spaces.

It just surround the path variable with in double quote in the string literal.

(This is my first public PR, I've looked for a contribution guide but did not find it... If there is anything wrong with the way I made it feel free to discard it and fix the concerned line directly, I'll learn by watching :wink: )

### DIFF
--- a/src/commands/init.ts
+++ b/src/commands/init.ts
@@ -147,8 +147,8 @@ export default defineCommand({
     }
     if (ctx.args.gitInit) {
       consola.info('Initializing git repository...\n')
-      const { execaCommand } = await import('execa')
-      await execaCommand(`git init ${template.dir}`, {
+      const { execa } = await import('execa')
+      await execa('git', ['init', template.dir], {
         stdio: 'inherit',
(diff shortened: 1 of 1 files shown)


src/commands/init.ts
```diff
@@ -148,7 +148,7 @@ export default defineCommand({
     if (ctx.args.gitInit) {
       consola.info('Initializing git repository...\n')
       const { execaCommand } = await import('execa')
-      await execaCommand(`git init ${template.dir}`, {
+      await execaCommand(`git init "${template.dir}"`, {
```
danielroe, line 151: Maybe worth using `execa` directly so it can escape the argument for us?

Ettapp:
Not sure I did the right thing by force-pushing the suggested change, but here it is, thanks for the advice !

pi0 (approved):
Thanks!

PR #302, opened 2023-11-29
feat(build): export shared utils

stafyniaksacha:
Follow https://github.com/nuxtlabs/nuxt-component-meta/pull/60#discussion_r1391400165

This PR exports internal utils, using mkdist, to allow creating custom CLI commands.

I also updated the `checkEngines` arguments to make it more reusable.


build.config.ts
```diff
@@ -8,7 +8,14 @@ export default defineBuildConfig({
       exportConditions: ['production', 'node'] as any,
     },
   },
-  entries: ['src/index'],
+  entries: [
+    'src/index', 
+    {
+      builder: 'mkdist',
```
pi0, line 14: Since utils should be tree-shakable it would be the best if we use a bundled entity. It also assure we can reuse same code and keep bundle size minimal

package.json
```diff
@@ -7,7 +7,63 @@
...
+    "./utils/packageManagers": {
+      "types": "./dist/utils/packageManagers.d.ts",
+      "import": "./dist/utils/packageManagers.mjs"
+    },
+    "./utils/templates": {
+      "types": "./dist/utils/templates.d.ts",
+      "import": "./dist/utils/templates.mjs"
+    },
+    "./utils/update": {
+      "types": "./dist/utils/update.d.ts",
```
pi0, line 64: Let's export utils by usage. Things like CJS utils and package manager utils also fs are so internal. I expect we export utils from cli that makes sense and having stable API without possibility of breaking changes.

PR #307, opened 2023-12-20
fix: Fix the template cannot be downloaded when the command 'nuxi ini…

GuoJikun:


src/commands/init.ts
```diff
@@ -10,7 +10,7 @@ import { defineCommand } from 'citty'
 import { sharedArgs } from './_shared'
 
 const DEFAULT_REGISTRY =
-  'https://raw.githubusercontent.com/nuxt/starter/templates/templates'
```
pi0, line 13: This seems to b e working:

https://raw.githubusercontent.com/nuxt/starter/templates/templates/v3.json

pi0: (while new change, with default branch, not https://raw.githubusercontent.com/nuxt/starter/templates/v3.json)

GuoJikun: You are right, but I always get a template download error when executing the 'nuxi init' command, which can only be solved by configuring the template address by the environment variable itself.

pi0: Perhaps you get any issue within proxy?

GuoJikun: > Perhaps you get any issue within proxy?

maybe
In theory, the url to download the template from the first and third terminals should be the same. So I'm a little confused.

pi0: It could be also a cached version of (blocked) github response. Can you message me in discord (@ pi0) would love to check together about this issue.

Issue #343, opened 2024-02-18
Nuxi doesn't work with volta.


tmg0:
### Environment

volta: 1.1.0

### Reproduction

https://github.com/zekunjin/volta-node-version

### Describe

I used volta in local environment, after pin a node version in `package.json`, nuxi does not use it, so I have to use `volta run` to declare a node version.

I'm not sure if it's possible to fetch the current project's Node.js version when executing the nuxi command, and then run it with that specified version without use `volta run`.

PR #344, opened 2024-02-18
feat: support volta

tmg0:
### Linked issue

Resolves: #343 

### Describe

Do not need to use `volta run` to declare a specified version of Node.js.

Use `execa` to run `nuxi.mjs` with a volta wrapper.

First check if there has `volta` command in local environment, and then use `volta which node` check which node version will be use in current project, finally transfer the command from `node ./bin/nuxt.mjs` -> `volta run --node 18.0.0 node ./bin/nuxt.mjs`

It might also be possible to add some logic like if the local environment has Volta and the `package.json` does not specify a Node version, `nuxi` could specify a Node version to run.

Resolves: #343

### DIFF
--- a/bin/_.mjs
+++ b/bin/_.mjs
@@ -1,0 +1,11 @@
+#!/usr/bin/env node
+
+import { fileURLToPath } from 'node:url'
+import { runMain } from '../dist/index.mjs'
+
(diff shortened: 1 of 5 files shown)


tmg0:
Thank you for the explanation about this issue

PR #362, opened 2024-03-15
feat: support module aliases and validate cwd and pkg

Atinux:
Add support to install module with alias:

CleanShot 2024-03-15 at 21 05 51@2x

See https://github.com/nuxt/nuxt/issues/26277 and https://github.com/nuxt/modules/commit/2376fd0538b9ddc6ed92feb531398f99a1e6ebfa

By looking at the issue, I also added a check to make sure that the package we install have at least Nuxt as dependency/devDependency instead of just installing an npm module (as some Nuxt modules are not part of nuxt/modules repository yet).

CleanShot 2024-03-15 at 21 05 39@2x

It also detects if `nuxt` is a dev dependency or not to install the module as dev dep or not.


README.md
```diff
@@ -14,6 +14,21 @@ We've rebuilt `nuxi` from the ground up using [unjs/citty](http://github.com/unj
...
+# Install dependencies
+pnpm i
+
+# Generate type stubs
+pnpm dev:prepare
+
+# Go the the playground directory
+cd playground
+# And run any commands
+pnpm nuxi <command>
```
Atinux, line 29:
```suggestion

# And run any commands
pnpm nuxi <command>
```

pi0 (approved):
🚀

src/commands/module/add.ts
```diff
@@ -211,10 +212,28 @@ async function resolveModule(
     }
   }
 
+  // Fetch package on npm
+  pkgVersion = pkgVersion || 'latest'
+  const pkg = await $fetch(`https://registry.npmjs.org/${pkgName}/${pkgVersion}`)
```
mianlang: hardcoding registry makes lives of those with poor network access to official registry harder

src/commands/module/add.ts
```diff
@@ -211,10 +234,33 @@ async function resolveModule(
     }
   }
 
+  // Fetch package on npm
+  pkgVersion = pkgVersion || 'latest'
+  const pkg = await $fetch(
+    `https://registry.npmjs.org/${pkgName}/${pkgVersion}`,
```
mianlang, line 240: hardcoding registry makes lives of those with poor network access to official registry harder

PR #370, opened 2024-03-20
fix(dev): prepend `baseURL` for vite hmr ws upgrade

antfu:
This was causing errors for apps with custom `baseUrl` and making dev server non-usable at all

Screenshot 2024-03-19 at 21 00 53

### DIFF
--- a/src/utils/dev.ts
+++ b/src/utils/dev.ts
@@ -245,9 +245,10 @@ class NuxtDevServer extends EventEmitter {
         'upgrade',
         async (req: any, socket: any, head: any) => {
           const nuxt = this._currentNuxt
-          if (!nuxt)
-            return
-          const viteHmrPath = (nuxt.options.app.baseURL + nuxt.options.app.buildAssetsDir).replace(/\/\//g, '/')
+          if (!nuxt) return
+          const viteHmrPath = (
+            nuxt.options.app.baseURL + nuxt.options.app.buildAssetsDir
+          ).replace(/\/\//g, '/')
           if (req.url.startsWith(viteHmrPath)) {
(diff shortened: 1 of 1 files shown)


src/utils/dev.ts
```diff
@@ -244,14 +244,15 @@ class NuxtDevServer extends EventEmitter {
...
-          if (
-            req.url.startsWith(
-              this._currentNuxt?.options.app.buildAssetsDir /* /_nuxt/ */,
-            )
-          ) {
+          const nuxt = this._currentNuxt
+          if (!nuxt) return
+          const viteHmrPath = (
+            nuxt.options.app.baseURL + nuxt.options.app.buildAssetsDir
+          ).replace(/\/\//g, '/')
```
antfu, line 251: You might have better method to combine two URLs - feel free to updates

pi0: Can we try ufo.joinURL?

src/utils/dev.ts
```diff
@@ -244,14 +244,15 @@ class NuxtDevServer extends EventEmitter {
...
+          if (req.url.startsWith(viteHmrPath)) {
             return // Skip for Vite HMR
           }
-          await this._currentNuxt?.server.upgrade(req, socket, head)
+          await nuxt.server.upgrade(req, socket, head)
```
pi0, line 255: Was this change also needed as part of fix or simply refactor?

antfu: It's a refactor to get rid of the `?` and also serves as a defensive guard to avoid we checking `undefined/undefined`

pi0 (approved):

Issue #375, opened 2024-03-27
preview command failed with nitro.output.dir


xjccc:
1. set `nitro.output`
```nuxt.config.ts
 output: {
    dir: 'node-server',
    publicDir: 'node-server'
  }
```
2. pnpm build
3. pnpm preview

`nitro.json` is under `node-server`, but `preview` path is `server/node-server`
```
// preview output path
[ 'Users/playground/server/node-server/nitro.json',
  'Users/playground/.output/nitro.json' ]
```

When i config `nitro.srcDir` is same output in `node-server`

pi0:
The issue should be solved via [nightly release channel](https://nuxt.com/docs/guide/going-further/nightly-release-channel)

PR #376, opened 2024-03-27
fix(preview): correct path with custom `nitro.output.dir`

xjccc:
https://github.com/nuxt/cli/issues/375

delete `nitro.srcDir`

### DIFF
--- a/src/commands/preview.ts
+++ b/src/commands/preview.ts
@@ -37,7 +37,6 @@ export default defineCommand({
 
     const resolvedOutputDir = resolve(
       config.srcDir || cwd,
-      config.nitro.srcDir || 'server',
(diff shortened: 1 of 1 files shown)


xjccc:
> Thanks. Have to locally doublecheck but this seems a nice fix. 
> 
> 
> 
> (for context, addition was from https://github.com/nuxt/nuxt/pull/18882)

Thanks for quick reply. I run the cli in my local. And set config with playground, find this simple way to resolve path. I don't know is something change build by nitro？

And i also have a question, when i set `~/node-server` the path can't resolve correct. But it's a special scene for this, it's need to solve？

pi0 (approved):
Thanks for PR and sorry for delay on merging it

Issue #361, opened 2024-03-15
When working with Layers, Nuxt adds the module to the root nuxt.config.ts but that is not what I wanted


martinszeltins:
### Describe the feature

When working with Nuxt Layers and you add a Nuxt module using the Nuxt Devtools or using nuxi `nuxi module add tailwind` it adds the module to the root `nuxt.config.ts` file and this is not configurable.

My `nuxt.config.ts` file is only for extending layers and I do not want Nuxt to add modules here. It would be very nice if this was configurable _(or if asked me which nuxt.config.ts to use)_ so that I can specify which `nuxt.config.ts` file I actually want it to use. Maybe before adding the module it could ask me which one I want it to add the module to.

````
common
  └─ components
  └─ composables
  └─ nuxt.config.ts   <------ I actually want nuxi to add the module here
users
  └─ components
  └─ composables
  └─ nuxt.config.ts
nuxt.config.ts        <--------- it adds it here (root)
````

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

PR #380, opened 2024-03-29
feat(module): allow adding module to Nuxt config within layer

DamianGlowala:
resolves https://github.com/nuxt/cli/issues/361

This allows for adding a module entry to a Nuxt config file within a layer using `--layer <layer_name>`.

### DIFF
--- a/src/commands/module/add.ts
+++ b/src/commands/module/add.ts
@@ -35,6 +35,10 @@ export default defineCommand({
       type: 'boolean',
       description: 'Skip nuxt.config.ts update',
     },
+    layer: {
+        type: 'string',
+        description: 'Layer name'
+    }
   },
   async setup(ctx) {
     const cwd = resolve(ctx.args.cwd || '.')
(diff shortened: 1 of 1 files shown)


src/commands/module/add.ts
```diff
@@ -88,7 +92,9 @@ export default defineCommand({
 
     // Update nuxt.config.ts
     if (!ctx.args.skipConfig) {
-      await updateNuxtConfig(cwd, (config) => {
+      const rootDir = ctx.args.layer ? resolve(cwd, ctx.args.layer) : cwd
```
pi0, line 95: Without considering layer names, isn’t it same as —cwd ?

DamianGlowala: I think setting `--cwd` pointing to the layer dir would affect other places as well, e.g. where `module add` looks for package.json or where it adds a dependency using `nypm` util. Correct me if I'm wrong though.

pi0: I think what is expected is that the module is added to the closest package.json and nuxt.config files found to the layer. (ie: We need a look-up search for both) 

I think it would probably be best first action to resolve any issues that prevents this. 

—-

In next steps, —layer would be also good idea but i expect that it loads nuxt.config, and tries to match a layer by its name (which used in extends key) basically as a convenient shortcut and also it should implement safety check that we don’t write to readonly (remote or dependency) layers but instead ask to directly add to the root.

PR #381, opened 2024-03-29
fix(dev,preview,typecheck): use explicit `--dotenv` filename

DamianGlowala:
resolves https://github.com/nuxt/cli/issues/360

This corrects types for the `--dotenv` flag and sets the default value at source.


danielroe (commented):
Will running `npx nuxi preview --dotenv` continue to work as expected (ie. a non-breaking change)?

src/commands/preview.ts
```diff
@@ -89,14 +89,22 @@ export default defineCommand({
...
+      if (envExists) {
+        logger.info(
+          `Loading \`${envFileName}\`. This will not be loaded when running the server in production.`,
+        )
+
+        await setupDotenv({ cwd, fileName: envFileName })
+      }
+      else {
+        logger.error(`Cannot find \`${envFileName}\`.`)
+        process.exit(1)
```
danielroe, line 106: should we just warn in this case?

danielroe (commented):
checking, this doesn't seem to resolved the linked issue - `.env` is used as a the 'fallback' instead of `.env.local`.

danielroe (approved):

PR #412, opened 2024-05-16
chore: migrate to eslint v9

danielroe:
closes https://github.com/nuxt/cli/pull/409

this implements a plain/vanilla nuxt eslint approach. removed prettier in favour of the stylistic approach adopted by anthony, but feel free to update/revert rule changes as you like!


pi0 (approved):
(consistency is most valuable, thanks!)

PR #422, opened 2024-05-30
feat(init): Use user provided dir for package.json name field

### LINKED ISSUE #421: use project name for package.json name in init Command
## Description

When running the command `npx nuxi@latest init <project-name>`, the `package.json`'s `name` is always initialized with `nuxt-app`
I am aware that we are fetching the files from the templates but do you think that would be possible to make it dynamic to match the `<project-name>`?

## usecase
This enhancement would improve consistency and streamline the initialization process for new Nuxt projects specially in monorepo projects I'd say.

Eckhardt-D:
Closes #421 

Causes some perf overhead due to 2 syscalls to read/write pkg. Can be addressed by adding source transform capabilities in `giget` before writing to disk.


danielroe (approved):
so sorry this took so long!

Issue #357, opened 2024-03-10
Run nuxt with different `envName` for env-specific configs


manniL:
Hey all 👋 

[c12](https://github.com/unjs/c12?tab=readme-ov-file#environment-specific-configuration) provides us with the amazing opportunity to add env-specific configs to the nuxt config (also [covered in my video](https://www.youtube.com/watch?v=DFZI2iVCrNc)), therefore relying on `NODE_ENV` by default. While it is technically possible to change the `envName` value, e.g. by patching the CLI or similar, it would be nice to have some kind of flag which changes the `envName` to a different env variable, such as `TARGET`, `APP_ENV` etc.

pi0:
Can you share more context in what common usecases this might be useful to use something different from `NODE_ENV`?

manniL:
In addition to @fmoessle's comment, `build` sets the NODE_ENV always to `production`, leaving no room for further options.

pi0:
> In addition to @fmoessle's comment, build sets the NODE_ENV always to production, leaving no room for further options.

This is a bug we might try to fix first.

fmoessle:
I think not using `process.env.NODE_ENV` with ```production``` when building the nuxt application is not a good idea. There are a lot of packages that behave differently when `process.env.NODE_ENV` is set to ```production``` (e.g. [Node docs](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production)).

So as @manniL pointed out, `process.env.NODE_ENV` should not be used here.

pi0:
> I think this is intentional, isn't it? 👀

It is intentionally set because tools (notably postcss stuff) hardly depend on `NODE_ENV==='production'` to optimize properly. 

It is not intentional that we don't respect it for c12 configuration env.

(@fmoessle  for clarification I made that decision as well ;)

maximepvrt:
And allow to use [environment overrides](https://nuxt.com/docs/getting-started/configuration#environment-overrides) for all custom environments (ex: staging)

tillsanders:
This would be a great help in supporting more than two environments. We, for example, have an additional demo environment for sales and a public beta environment that need differentiation (e.g. when it comes to CSP or the title template).

I also have a question though: would the environment-specific configuration be 'backed in' during build time or would it still be possible to change the environment when starting the build, for example like this: `CUSTOM_ENV=demo node ./output/server/index.mjs`?

tillsanders:
[...]

PR #424, opened 2024-05-31
feat: add env-flag for env-specific config (#357)

tillsanders:
This PR is proposing to add a new `--env [name]` attribute to Nuxi. The attribute will be passed to the `envName` option of the c12 library. This adds the ability to support environment overrides (environment-specific configuration) for custom environments. Currently, c12 supports that and the types of the nuxt configuration file are aware of that, but there is no way to actually use custom environments, because `NODE_ENV` cannot be set to values other than `production` or `development`. This PR closes the gap, making deployments to different environments easier.

See #357.

Should this be approved, I would be happy to amend the documentation accordingly.

### DIFF
--- a/src/commands/_shared.ts
+++ b/src/commands/_shared.ts
@@ -9,6 +9,13 @@ export const sharedArgs = {
   },
 } as const
 
+export const envNameArgs = {
+  envName: {
+    type: 'string',
+    description: 'Name of the build environment to use (see \'Environment overrides\' in the docs)',
+  },
+} as const
+
 export const legacyRootDirArgs = {
   rootDir: {
     type: 'positional',
--- a/src/commands/_shared.ts
+++ b/src/commands/_shared.ts
@@ -12,7 +12,7 @@ export const sharedArgs = {
 export const envNameArgs = {
   envName: {
     type: 'string',
-    description: 'Name of the build environment to use (see \'Environment overrides\' in the docs)',
+    description: 'The environment to use when resolving configuration overrides (default is `production` when building, and `development` when running the dev server)',
   },
 } as const
 
--- a/src/commands/build.ts
+++ b/src/commands/build.ts
@@ -27,6 +27,10 @@ export default defineCommand({
       type: 'string',
       description: 'Path to .env file',
     },
+    env: {
+      type: 'string',
+      description: "Name of the build environment to use (see 'Environment overrides' in the docs)",
+    },
     ...legacyRootDirArgs,
   },
   async run(ctx) {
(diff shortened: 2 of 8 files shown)


MickL:
Does this PR also apply for the environment configs set in `nuxt.config.ts`?

   $production: {
       ignore: ['app/pages/dev'],
       scripts: {
         registry: {
           cloudflareWebAnalytics: true,
         },
       },
   },
   $preview: {
        ignore: ['app/pages/dev'],
   },
});

src/commands/build.ts
```diff
@@ -27,6 +27,10 @@ export default defineCommand({
       type: 'string',
       description: 'Path to .env file',
     },
+    env: {
```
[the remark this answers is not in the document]
pi0: PS: i think we can make a shared object similar to how `legacyRootDirArgs` now i guess since this is arg is shared for all commands.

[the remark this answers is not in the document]
tillsanders: Thank you, I appreciate your feedback! I made the necessary changes as you suggested. I noticed that there is a slight inconsistency in the way the flags are named. Some use kebab-case, others lowerCamelCase – e.g. `--logLevel` and `--no-fork`. I decided to go with lowerCamelCase as that seems to be the majority and matches the naming in the `loadNuxt()` function. 

I also made a small amendment to the documentation, here: https://github.com/nuxt/nuxt/pull/28909

Please let me know if any other changes are required – looking forward to have this available!

tillsanders:
I added a commit to address the feedback above. Is there anything else I can do to improve this pull request?

src/commands/_shared.ts
```diff
@@ -9,6 +9,13 @@ export const sharedArgs = {
   },
 } as const
 
+export const envNameArgs = {
+  envName: {
+    type: 'string',
+    description: 'Name of the build environment to use (see \'Environment overrides\' in the docs)',
```
pi0, line 15: I would suggest something that is quickly understandable in usage

```suggestion
    description: 'Environment variable to use for resolving config (default is `NODE_ENV`)',
```

pi0 (approved):
other than usage, looks good to me

src/commands/_shared.ts
tillsanders: Agreed, however, the description you suggest does not match the implementation. The value passed to the  envName argument does not match an env variable but the environment name in the Nuxt configuration. This behaviour matches the behaviour of c12.

pi0: Indeed c12 does config resolution right? it is too much impl detail if we want to mention low level library IMO. Any better ideas is welcome!

tillsanders: In that case, I would suggest something like 'Name of the environment overrides to use for resolving config'. This uses the current wording of the Nuxt configuration, here: https://nuxt.com/docs/getting-started/configuration#environment-overrides

PR #433, opened 2024-06-23
feat: `add` command support for `.config` folder

### LINKED ISSUE #400: `add` command creates `nuxt.config.ts` while `nuxt.config.js` exists
Not sure if this expected, when running the `add` command (for example `npx nuxi@latest module add nuxt-viewport`) in a project using `nuxt.config.js` a new file is created `nuxt.config.ts` with the new module:
```
// https://nuxt.com/docs/api/configuration/nuxt-config
  modules: ["nuxt-viewport"]
})
```
I would expect it to add the new module to the exisiting config file.

Reproduce this issue on StackBlitz starter: https://stackblitz.com/fork/github/nuxt/starter/tree/v3-stackblitz

Apologies if this is an upstream issue.

Yizack:
Add support for `.config` folder files. Closes #400

- Forked from: #401


src/commands/module/add.ts
```diff
@@ -116,28 +116,42 @@ async function updateNuxtConfig(
...
-  const nuxtConfigFile = resolve(rootDir, 'nuxt.config.ts')
-  if (existsSync(nuxtConfigFile)) {
-    consola.info('Updating `nuxt.config.ts`')
-    _module = await loadFile(nuxtConfigFile)
+  const configFiles = [
+    'nuxt.config.ts',
+    'nuxt.config.js',
+    '.config/nuxt.ts',
+    '.config/nuxt.js',
+  ]
```
shinGangan, line 124: Hi @Yizack , I have a question.

### Ques1

Is it possible to detect when `.config/` are nested?

https://github.com/pi0/config-dir?tab=readme-ov-file#nesting

Issue #374, opened 2024-03-22
Proposal: init test setup with Nuxi


rinux55:
One thing I'm really missing from `nuxi` is a way to quickly initialise a test setup. 

My proposal is that after people have initialised a project with `nuxi init`, they can run a command like `nuxi add tests` or `nuxi tests init` which will start a small wizard to set up everything related to testing.

Running this command will:

* install `vitest`, `@nuxt/test-utils` `vitest-environment-nuxt`
* adds a `vitest.config.ts`
* Asks to add `@vue/test-utils` or `@testing-library/vue` for unit testing
* Asks to add `happy-dom` or `jsdom`
* Asks to add `playwright` for e2e testing
* Sets up `vitest.config` accordingly

I'd be down to contribute with a PR, but first I'm interested in some feedback! What do you think of this?

pi0:
Hi. I think it is a nice idea. I think we can move it forward with to directions:

1) A new `app-test` template in [nuxt/starter](https://github.com/nuxt/starter) that can initialize `/test` directory. Using `nuxi init -t test --instal` it can achieve (almost) same result
2) Later integrate it with upcoming scaffolding tooling to allow defining above steps without need of a template (old tracker: https://github.com/nuxt/cli/issues/57)

rinux55:
> * A new `app-test` template in [nuxt/starter](https://github.com/nuxt/starter) that can initialize `/test` directory. Using `nuxi init -t test --install` can achieve (almost) the same result

I am not super familiar with the starter templates, but will this allow for a wizard like approach like asking for `playwright` etc, or can this only create a project from a template without any options?

I'm assuming this allows to _only_ create a starter template with tests, and not a combination of module **and** tests? I feel that this is too limiting, as in the ideal world it should be easy to add tests anywhere in a nuxt project. For instance, I'm creating Layers often, it would be really nice to have one command take care of it all.

My vision is that testing Nuxt applications (and with that, setting up testing) should be just as easy and fun as it is to work with the other aspects of Nuxt. That is why your second suggestion:

> * Later integrate it with upcoming scaffolding tooling to allow defining the above steps without the need of a template (old tracker: https://github.com/nuxt/cli/issues/57)

sounds like the ideal solution. If there is any way I can help making this a reality, let's have a chat.

pi0:
> I am not super familiar with the starter templates, but will this allow for a wizard-like approach like asking for a playwright, etc, or can this only create a project from a template without any options?

[...]

PR #448, opened 2024-07-05
feat: setup wizard

### LINKED ISSUE #53: CLI Wizard
I really liked the cli wizard in nuxt2.
image
Will it return in nuxt3 at some point?

Lehoczky:
### 🔗 Linked issue

Resolves #53 
Resolves https://github.com/nuxt/cli/issues/374

### 📚 Description

This PR implements an initial setup wizard which can be extended later.

Before implementing this feature I looked at some other framework CLI's:
- https://github.com/angular/angular-cli
- https://github.com/withastro/astro/tree/main/packages/create-astro
- https://github.com/sveltejs/kit/tree/main/packages/create-svelte
- https://github.com/vuejs/create-vue

When it comes to rendering the templates, I chose to [follow the Angular CLI, and use template files](https://angular.dev/tools/cli/schematics-for-libraries#add-template-files). For templating engine, [nunjucks](https://mozilla.github.io/nunjucks/) was chosen, because it is mature and have a nice template syntax, but at the end, it was a personal choice. I'm open to changing it to something else if there is a reason to do that.

For the options I followed [create-svelte and ask opinionated questions](https://github.com/sveltejs/kit/blob/main/packages/create-svelte/bin.js#L85-L106). For example, the CLI does not present choices for E2E frameworks (cypress / playwright), instead asks whether the user wants to use Playwright for E2E testing.  
The benefit of this approach is that it mitigates [decision fatigue](https://en.wikipedia.org/wiki/Decision_fatigue) and makes the CLI more maintainable (fewer choices, less code). It also helps newcomers by choosing tools for them.

Features:
- the user can chose to use prettier for formatting
- eslint for linting
- playwright for e2e testing
- vitest for unit testing
- recommended vscode settings

Choosing playwright with eslint configures the [eslint playwright plugin](https://www.npmjs.com/package/eslint-plugin-playwright).
Choosing vitest with eslint configures the [eslint vitest plugin](https://www.npmjs.com/package/eslint-plugin-vitest).
Choosing vitest also sets up [@nuxt/test-utils](https://www.npmjs.com/package/@nuxt/test-utils) according to the Nuxt documentation.

Later I can add back the "UI Framework" choice which was present in Nuxt2 to allow automatically setting up some of the more popular UI libraries, like tailwindcss, unocss, primevue, vuetify.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist
 .nuxt
 nuxt-app
 .pnpm-store
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules
 dist
-.vscode
 .nuxt
 nuxt-app
 .pnpm-store
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,0 @@
-{
-  "semi": false,
-  "singleQuote": true,
-  "vueIndentScriptAndStyle": false
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "prettier.enable": false
+}
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -4,12 +4,10 @@ import { createConfigForNuxt } from '@nuxt/eslint-config/flat'
 export default createConfigForNuxt({
   features: {
     tooling: true,
-    stylistic: true,
+    stylistic: false,
   },
   dirs: {
-    src: [
-      './playground',
-    ],
+    src: ['./playground'],
   },
 }).append({
   rules: {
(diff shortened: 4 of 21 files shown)


Lehoczky:
I reeeeally don't like pinging maintainers, and I know Pi is very busy with other projects, so I definitely won't ping him here, but @atinux I saw you commenting on other issues, maybe you or someone else from the team could look at this PR?

Just asking because I see sometimes people asking questions about setting up linting, or testing, and this PR could help them a lot I think. It would also close one of the most liked tickets in the repository. If somethings needs to be changed, I'm happy to do it!

pi0:
Thanks for your nice works here dear @Lehoczky i am a fan of nunjucks and we used a similar templating system for nuxt<3. 

For moving forward with an ideal wizard system, we need to support it at [unjs/giget](https://github.com/unjs/giget) layer and i made some (very early) attempts to add peer utils ([unjs/codeup](https://github.com/unjs/codeup)).

Please don't hesitate to ping me in discord if interested to collaborate on this. ❤️ 

/cc @danielroe marking this PR as draft for clarity but I'm on it and if for any reason you were thinking we need to speedup progress on this please ping me 🙏🏼

PR #449, opened 2024-07-06
fix: use -w arg for pnpm root workspace

huang-julien:
Hi :wave: 

This PR adds `-w` arg to the install command when having pnpm and pnpm workspaces. Pnpm prevents running installing new deps in the root directory without `-w`

### DIFF
--- a/src/commands/upgrade.ts
+++ b/src/commands/upgrade.ts
@@ -1,4 +1,5 @@
 import { execSync } from 'node:child_process'
+import { existsSync } from 'node:fs'
 import { consola } from 'consola'
 import { colors } from 'consola/utils'
 import { relative, resolve } from 'pathe'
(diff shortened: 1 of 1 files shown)


src/commands/upgrade.ts
```diff
@@ -88,7 +94,7 @@ export default defineCommand({
     execSync(
       `${packageManager} ${
         packageManager === 'yarn' ? 'add' : 'install'
-      } -D nuxt`,
+      } -D nuxt ${packageManager === 'pnpm' && hasPnpmWorkspaceFile() ? '-w' : ''}`,
```
danielroe, line 97:
```suggestion
      } -D nuxt ${packageManager === 'pnpm' && hasPnpmWorkspaceFile(cwd) ? '-w' : ''}`,
```

src/commands/upgrade.ts
```diff
@@ -27,6 +28,11 @@ async function getNuxtVersion(path: string): Promise<string | null> {
   }
 }
 
+function hasPnpmWorkspaceFile(): boolean {
+  const pnpmWorkspaceFilePath = `${process.cwd()}/pnpm-workspace.yaml`
+  return existsSync(pnpmWorkspaceFilePath)
+}
```
danielroe, line 34:
```suggestion
function hasPnpmWorkspaceFile(cwd: string): boolean {
  const pnpmWorkspaceFilePath = join(cwd, 'pnpm-workspace.yaml')
  return existsSync(pnpmWorkspaceFilePath)
}
```

danielroe (approved):

Issue #454, opened 2024-07-30
`nuxi module add` always install module as dependcies, even it should be `devDependencies`


gxres042:
For example, when I install `@nuxt/eslint`, for pnpm, I will install `eslint` and `@nuxt/eslint` as `devDepencies`:

``` shell
pnpm add -D @nuxt/eslint eslint
```

But when I use `nuxi module add eslint`, it runs:

``` shell
pnpm add @nuxt/eslint // didn't have eslint
```

Atinux:
Actually this is how we detect where to install the dependency: https://github.com/nuxt/cli/blob/bca75ab6490dfc444e3b6766ba5c47572622e2d3/src/commands/module/add.ts#L63

I think we can add an exception for `@nuxt/eslint` to force being a dev dependency anyway

Happy to open a PR for it?

gxres042:
Maybe try it 🥰

gxres042:
@Atinux I found a question

When initializing a Nuxt 3 project, nuxt and vue are installed by default as dependencies instead of devDependencies. The detection above is done by checking if nuxt is a devDependency, i.e. if nuxt in the user's project is not a devDependency, then any dependencies added by the user's module add will not be installed as devDependencies.

I'm not quite sure if all the official Nuxt modules need to be installed as devDependencies, but what I do know so far is that the manual install of @nuxt/eslint requires it and eslint to be installed as devDependencies.

atinux:
Actually, everything can be a `devDependency` in Nuxt 3 as we bundle everything for production (including the depdendencies used) :)

PR #463, opened 2024-08-25
feat(module): Force new dependencies to be devDependencies (if it's in a specific list)

gxres042:
### 🔗 Linked issue

#454 

### ❓ Type of change

### 📚 Description

This PR makes some changes to the module add directive: it detects a specific list of npm dependencies when adding them, and if the npm dependency to be installed is in that list, it will be installed as a development dependency.

While Nuxt will extract what is needed into the final build product, dependencies like `@nuxt/eslint` should not be installed as “dependencies” but as “devDependencies” in the original meaning of dependencies and devDependencies. Hence the PR.

It's not a good practice to write a dead list inside the module add function, but I haven't come up with a perfect way to do it at this stage either.

### DIFF
--- a/src/commands/module/add.ts
+++ b/src/commands/module/add.ts
@@ -15,6 +15,12 @@ import {
 } from './_utils'
 import type { NuxtModule } from './_utils'
 
+// Maybe we can import the list from somewhere, but idk
+const forceDevDependencies = [
+  "@nuxt/eslint@latest",
+  "@nuxtjs/color-mode"
+]
+
 export default defineCommand({
   meta: {
     name: 'add',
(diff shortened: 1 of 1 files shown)


src/commands/module/add.ts
```diff
@@ -15,6 +15,12 @@ import {
 } from './_utils'
 import type { NuxtModule } from './_utils'
 
+// Maybe we can import the list from somewhere, but idk
```
danielroe, line 18: in general I would think modules should match `nuxt` rather than forcing modules to development mode. they will be required in the same environment as nuxt (i.e the build environment).

if anything we could support a flag to allow the user to choose (`--dev`).

pi0: (we could also add this list to the module registry meta btw)

danielroe:
I think we can mark this implemented in https://github.com/nuxt/cli/pull/630.

PR #466, opened 2024-09-03
DRAFT: Allow reloading of env variables before restarting nuxt dev server

abhi12299:
### 🔗 Linked issue

Refer to https://github.com/nuxt/nuxt/issues/28543 to reproduce this issue

### ❓ Type of change

### 📚 Description

When the contents of `.env` change, the dev server is unable to pick up the changes, even after the server restarts.

There are multiple ways to go about implementing this feature, one of which is demonstrated in this PR. `c12` library is used to setup and load environment variables from `.env` file. The following code is responsible for the same:
https://github.com/nuxt/cli/blob/34bdbbcbee9612696fd9cecf0de0ffc3421cc744/src/commands/dev.ts#L47-L52

The function responsible for restarting the child process is:
https://github.com/nuxt/cli/blob/34bdbbcbee9612696fd9cecf0de0ffc3421cc744/src/commands/dev.ts#L167-L187

You'll notice that `process.env` is expanded directly, and there is no step to reload the environment variables from the changed `.env` file. 

The change I proposed keeps track of the `cwd` and `dotenv` args passed to the `dev` command and then run this code prior to forking a new process:
```ts
if (_cwd) {
    const env = await loadDotenv({ cwd: _cwd, fileName: _dotenv || '.env', interpolate: true })
    Object.assign(process.env, env)
}
```

The reason for not using the same `setupDotenv` function provided by `c12` is because if you look at the definition of the function (taken from `c12`'s repo):
```ts
export async function setupDotenv(options: DotenvOptions): Promise<Env> {
  const targetEnvironment = options.env ?? process.env;

  // Load env
  const environment = await loadDotenv({
    cwd: options.cwd,
    fileName: options.fileName ?? ".env",
    env: targetEnvironment,
    interpolate: options.interpolate ?? true,
  });

  // Fill process.env
  for (const key in environment) {
    if (!key.startsWith("_") && targetEnvironment[key] === undefined) { // -> ***notice this line
      targetEnvironment[key] = environment[key];
    }
  }

  return environment;
}

/** Load environment variables into an object. */
export async function loadDotenv(options: DotenvOptions): Promise<Env> {
  const environment = Object.create(null);

  const dotenvFile = resolve(options.cwd, options.fileName!);

  if (existsSync(dotenvFile)) {
    const parsed = dotenv.parse(await fsp.readFile(dotenvFile, "utf8"));
    Object.assign(environment, parsed);
  }

  // Apply process.env
  if (!options.env?._applied) {
    Object.assign(environment, options.env);
    environment._applied = true;
  }

  // Interpolate env
  if (options.interpolate) {
    interpolate(environment);
  }

  return environment;
}
```

Assume that an environment variable `NUXT_API_KEY` gets changed. Now the line marked with `***` will check if the key is undefined in the target environment, i.e. `process.env`, which in this case is actually defined. So it will never overwrite the value.

*I've marked this PR as draft because we can instead request upstream changes to `c12` to tidy up the code here.*

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -2,7 +2,7 @@ import { fork } from 'node:child_process'
 import type { ChildProcess } from 'node:child_process'
 import type { IncomingMessage, ServerResponse } from 'node:http'
 import { resolve } from 'pathe'
-import { setupDotenv } from 'c12'
+import { loadDotenv, setupDotenv } from 'c12'
 import type { ParsedArgs } from 'citty'
 import { defineCommand } from 'citty'
 import { isBun, isTest } from 'std-env'
@@ -20,6 +20,8 @@ import type { NuxtDevContext, NuxtDevIPCMessage } from '../utils/dev'
 import { sharedArgs, legacyRootDirArgs } from './_shared'
 
 const forkSupported = !isBun && !isTest
+let _cwd: string
+let _dotenv: string | undefined
 
 const command = defineCommand({
   meta: {
@@ -48,6 +50,8 @@ const command = defineCommand({
     // Prepare
     overrideEnv('development')
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir || '.')
+    _cwd = cwd
+    _dotenv = ctx.args.dotenv
     showVersions(cwd)
     await setupDotenv({ cwd, fileName: ctx.args.dotenv })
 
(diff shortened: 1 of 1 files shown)


abhi12299:
@pi0 can you please review this? Would like to know your thoughts.

abhi12299:
@danielroe can you please review when you get time. I can work on an upstream fix if needed.

Issue #453, opened 2024-07-24
nuxi always using default npm registry


wdyyy:
### Environment

Nuxt project info:                                                                                                                                                             9:52:35 PM

------------------------------
- Operating System: Linux
- Node Version:     v20.12.2
- Nuxt Version:     3.12.4
- CLI Version:      3.12.0
- Nitro Version:    2.9.7
- Package Manager:  yarn@1.22.22
- Builder:          -
- User Config:      devServer, devtools, modules
- Runtime Modules:  @nuxt/ui@2.17.0, nuxt-quasar-ui@2.1.3
- Build Modules:    -
------------------------------

### Reproduction

Only in the region where users couldn't access https://registry.npmjs.org, like china, you can reproduce it.

### Describe the bug

No matter what registry I set, when I run npx nuxi@latest module add XXX, nuxi report errors like this:

```plaintext
ERROR  [GET] "https://registry.npmjs.org/@nuxt/ui/latest": <no response> fetch failed                                                                                         9:43:28 PM

  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async $fetchRaw2 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.ec1423ac.mjs:267:14)
  at async $fetch2 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.ec1423ac.mjs:304:15)
  at async resolveModule (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/chunks/add2.mjs:26359:15)
  at async Object.setup (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/chunks/add2.mjs:26219:15)
  at async runCommand$1 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.6aad497e.mjs:1620:5)
  at async runCommand$1 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.6aad497e.mjs:1639:11)
  at async runCommand$1 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.6aad497e.mjs:1639:11)
  at async runMain$1 (/home/wdy/.npm/_npx/b95349761371180e/node_modules/nuxi/dist/shared/nuxi.6aad497e.mjs:1777:7) 
```

[...]

PR #470, opened 2024-09-05
feat: add user-specified registry support (from npmrc)

Kiameow:
### 🔗 Linked issue

[nuxi always using default npm registry #453](https://github.com/nuxt/cli/issues/453)
### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality)

### 📚 Description

For user who doesn't have the stable connection to the default npm regsitry, they tend to use the mirror site, but nuxi doesn't have very good support about it (user can set COREPACK_NPM_REGISTRY env to achieve the ideal result, but at least for me, I didn't realize it at first time). 

Now the using of nuxi to install a package will directly read the registry info from user's .npmrc file, then change the registry to get package json info as well as the COREPACK_NPM_REGISTRY (if it isn't been set) to download the package


src/commands/module/add.ts
```diff
@@ -248,3 +253,19 @@ async function resolveModule(
     pkgVersion,
   }
 }
+
+function getRegistryFromNpmrc() {
+  const npmrcPath = join(homedir(), '.npmrc')
```
danielroe, line 258: should we also support the current directory?

Kiameow: sure, sir, support added

atinux:
Did you think about using https://www.npmjs.com/package/npm-registry-fetch ?

pi0:
> [s.com/package/npm-registry-fetch](https://www.npmjs.com/package/npm-registry-fetch) ?

this looks nice however has lots of ties with Node.js (also not sure if it respects corepack env, for reasons) I would personally pick simplest approach (can later refactor it out to nypm for easier maintenance)

danielroe (approved):
I think we can go ahead and merge as a basic implementation and later improve...

PR #522, opened 2024-10-18
Ensure the generated composable name is properly formatted

franklin-tina:
### ❓ Type of change

### 📚 Description
The current implementation assumes users will run the command to generate a composable as follows:
1. prefixed with use
`nuxi add composable useSomeComposable`

```ts
// useSomeComposable.ts
export const useSomeComposable = () => {
  return ref()
}
```

2. without `use` prefix
`nuxi add composable someComposable`

```ts
// someComposable.ts
export const useSomeComposable = () => {
  return ref()
}
```

However, if the user runs these commands with the names in kebab case, the resulting function will be broken:
1. prefixed with use
`nuxi add composable use-some-composable`

```ts
// use-some-composable.ts
export const use-some-composable = () => {
  return ref()
}
```

2. without `use` prefix
`nuxi add composable some-composable`

```ts
// some-composable.ts
export const useSome-composable = () => {
  return ref()
}
```

This PR ensures the resulting composable function is properly formatted, regardless of whether or not the user uses kebab case or camelCase in their composable definition:
```
use-some-composable → useSomeComposable
good → useGood
someGoodComposable → useSomeGoodComposable
usegood-composable → useGoodComposable
```


danielroe (approved):
thank you ❤️

PR #587, opened 2024-12-06
fix: `analyze` should include original readable filenames

### LINKED ISSUE #390: Analyze command should show descriptive file names
Hey! After https://github.com/nuxt/nuxt/pull/26203 landed, `nuxi analyze` suffers from less clear file names. As a workaround, I commonly add a snippet to remove the hash fully and use the name only (`[name].js`). It'd be great if analyze could do that by default for us though.

BobbieGoede:
### 🔗 Linked issue
* #390 

### ❓ Type of change

### 📚 Description
* Resolves #390 

This changes the `analyze` command to change the chunk filenames to include the original human readable filename. 

~~@manniL 
In the issue you mention removing the hash entirely, if that's preferable I can change the PR. I figured including the hash could prevent duplicate filenames (not sure if possible) and in case these are used to invalidate build cache (I have no idea 🤷)~~

### DIFF
--- a/src/commands/analyze.ts
+++ b/src/commands/analyze.ts
@@ -52,8 +52,8 @@ export default defineCommand({
           build: {
             rollupOptions: {
               output: {
-                chunkFileNames: '_nuxt/[name].[hash].js',
-                entryFileNames: '_nuxt/[name].[hash].js',
+                chunkFileNames: '_nuxt/[name].js',
+                entryFileNames: '_nuxt/[name].js',
               },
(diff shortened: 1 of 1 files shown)


danielroe:
I think the `[name]` would be fine on its own - rollup ensures it doesn't create duplicates.

BobbieGoede:
Nice, I removed the hash from the name pattern 🚀

danielroe (approved):

PR #590, opened 2024-12-06
fix(cleanup): remove dev buildDir, not production

### LINKED ISSUE #533: Nuxi Cleanup does not seem to clear Nuxt files in v4
### Environment

```
Nuxt project info:                                                                                                                              

------------------------------
- Operating System: Linux
- Node Version:     v18.20.3
- Nuxt Version:     3.13.2
- CLI Version:      3.14.0
- Nitro Version:    2.9.7
- Package Manager:  npm@10.2.3
- Builder:          -
- User Config:      -
- Runtime Modules:  -
- Build Modules:    -
------------------------------
```

### Reproduction

https://stackblitz.com/edit/nuxt-starter-tqwwsx

### Describe the bug

Hey Nuxt Team!

I recently upgraded some projects that to Nuxt4 using `compatibilityVersion: 4`. Everything works great, but I notice that `npx nuxi cleanup` does not clear any Nuxt files (like the `.nuxt` file) when running this command.

Perhaps I am doing this wrong, but even in a minimal v4 environment like the example above, the cleanup command does not seem to work properly.

### Additional context

After further testing, the `node_modules` cache data clears correctly. However, the `.nuxt` file _does not_. 

### Logs

_No response_

BobbieGoede:
### 🔗 Linked issue
* #533 

### ❓ Type of change

### 📚 Description
Resolves #533 

It seems like, after running the dev server when using `compatibilityVersion: 4`, loading the nuxt config will no longer give the configured `buildDir` which will instead be set to something like `/node_modules/.cache/nuxt/.nuxt`. This is not very useful as we're already clearing the `/node_modules/.cache` directory with this command.

I'm not entirely sure what causes this change, but I did find that passing `overrides: { dev: true }` to `loadNuxtConfig` returns the expected values, not sure if this has other unintended consequences, we're only using the `rootDir` and `buildDir` from the loaded config.

### DIFF
--- a/src/commands/cleanup.ts
+++ b/src/commands/cleanup.ts
@@ -17,7 +17,7 @@ export default defineCommand({
   async run(ctx) {
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir || '.')
     const { loadNuxtConfig } = await loadKit(cwd)
-    const nuxtOptions = await loadNuxtConfig({ cwd })
+    const nuxtOptions = await loadNuxtConfig({ cwd, overrides: { dev: true } })
     await cleanupNuxtDirs(nuxtOptions.rootDir, nuxtOptions.buildDir)
   },
 })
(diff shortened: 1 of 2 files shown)


danielroe (approved):

danielroe:
(In v4 mode, the buildDir is actually located in that cache folder so multiple builds can happen in parallel with a dev server.)

Issue #450, opened 2024-07-08
Plan for v4?


Barbapapazes:
Hello 👋,

Je me demandais s'il était prévu de passer à la version 4 de l'interface de programmation et de profiter de ce changement majeur pour retirer tout le legacy et rendre le CLI plus propre?

pi0:
Kit only has peer dependency on kit so if nuxi is used on nuxt v4 projects it will leverage it (including new jiti-esm resolutions) and we will eventually backport jiti v2 to the core of CLI (as non breaking change).

What are legacy things to make cleaner?

Barbapapazes:
Mainly talking about some args marked as legacy

For example:

https://github.com/nuxt/cli/blob/main/src/commands/build.ts#L30

pi0:
Alright, i've renamed issue to track specific drop. If you are thinking of more public interface breaking changes worth to consider please feel free to open new issue so we can have them in plan for v4 when it comes (personally i think just for one arg, it probably doesn't worth to make a major version because it fragments users only). We perhaps make v4 when essentially need to break things for Nuxt 3 users.

PR #592, opened 2024-12-08
refactor: shared command descriptions and `--cwd` default

BobbieGoede:
### 🔗 Linked issue
* #450

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation or readme)
- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### 📚 Description
Is this a refactor and/or docs change 🤷

I found the `ROOTDIR` and `cwd` default value or fallback value a bit confusing, these changes _should_ be functionally the same but instead handles the fallback to the shared argument configuration. Please make sure to check I'm not misunderstanding the fallback logic 😬

I'm slowly working on syncing the descriptions between the command usage `-h` output and those described on `docs/api/commands` pages, more PRs to come.

### DIFF
--- a/src/commands/_shared.ts
+++ b/src/commands/_shared.ts
@@ -1,32 +1,48 @@
-export const sharedArgs = {
+import type { ArgDef } from 'citty'
+
+export const cwdArgs = {
   cwd: {
     type: 'string',
-    description: 'Current working directory',
+    description: 'Specify the working directory, defaults to current directory (".")',
+    valueHint: 'directory',
+    default: '.',
   },
+} as const satisfies Record<string, ArgDef>
+
+export const sharedArgs = {
+  ...cwdArgs,
   logLevel: {
     type: 'string',
-    description: 'Log level',
+    description: 'Specify build-time log level.',
+    valueHint: 'silent|info|verbose',
   },
-} as const
+} as const satisfies Record<string, ArgDef>
 
 export const envNameArgs = {
   envName: {
     type: 'string',
     description: 'The environment to use when resolving configuration overrides (default is `production` when building, and `development` when running the dev server)',
   },
-} as const
+} as const satisfies Record<string, ArgDef>
 
 export const dotEnvArgs = {
   dotenv: {
     type: 'string',
-    description: 'Path to .env file',
+    description: 'Path to `.env` file to load, relative to the root directory.',
   },
-} as const
+} as const satisfies Record<string, ArgDef>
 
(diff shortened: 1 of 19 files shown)


BobbieGoede:
I split up `sharedArgs` into `cwdArgs` and `logLevelArgs`, as not all commands actually used the `--log-level`, this removes some noise from the docs as well in nuxt/nuxt#30199

src/commands/_shared.ts
```diff
@@ -1,32 +1,47 @@
...
+  // cwd falls back to rootDir's default (indirect default) to ease migration
+  cwd: {
+    ...cwdArgs.cwd,
+    description: 'Specify the working directory (default: `.`)',
+    default: undefined,
+  },
   rootDir: {
     type: 'positional',
-    description: 'Root Directory',
+    description: '(DEPRECATED) Use `--cwd` instead. Specifies the working directory (default: `.`)',
```
danielroe, line 43: hm.... my understanding was that the positional rootdir wasn't deprecated, it's the argument `--rootDir` that is deprecated.

the ROOTDIR here would be something like:

```bash
npx nuxi dev nuxt-app
```

but maybe I'm wrong - @pi0 were you thinking to deprecate the positional rootDir?

BobbieGoede: All I know is that `rootDir` was marked as legacy and that `---cwd` takes precedence over its usage since https://github.com/nuxt/cli/pull/47, while it wasn't explicitly marked as deprecated, dropping the argument was discussed in https://github.com/nuxt/cli/issues/450. 

I guess the main advantage of `--cwd` is that it isn't positional, but I can imagine there's plenty of scripts that rely on passing a `rootDir` directory without flag.

danielroe: Let's not deprecate the positional rootDir for now. 🙏

(I think the thing that's deprecated is the _name_ `rootDir` - but we can follow up on that in a separate PR.)

pi0: Yep, that was purpose of deprecation rootDir -> cwd by name. 

I wish we never had positional dir but that ship already sailed for nuxi main commands.

BobbieGoede: I see, I have removed the deprecation description now, but did add a description that `cwd` argument takes precedence over `rootDir` for clarity.

danielroe: perfect! ❤️

danielroe (approved):

PR #602, opened 2024-12-18
fix(dev): restart nuxt if unhandled rejection occurs

danielroe:
### ❓ Type of change

### 📚 Description

investigating https://github.com/nuxt/nuxt/issues/30137#issuecomment-2538199836, every time I started Nuxt, a zombie process would be left behind due to an unhandled rejection.

this adds handling to kill + restart nuxt, which doesn't address the issue (still investigating - it's about watchers) but at least prevents zombie processes from accumulating

(alternatively, or perhaps in future, we could add an additional signal if we wanted to handle differently - would love ideas @pi0)


src/commands/dev-child.ts
```diff
@@ -62,6 +50,24 @@ export default defineCommand({
         )
       }
     }
+
+    process.once('unhandledRejection', () => {
+      sendIPCMessage({ type: 'nuxt:internal:dev:restart' })
```
pi0, line 55: Perhaps we can show a clear warning about this behavior?

It looks a positive thing against unhandledRejections but also looks silent

PR #632, opened 2025-01-05
feat: use `jiti` to import/resolve modules

danielroe:
### ❓ Type of change

### 📚 Description

this drops a lot of legacy code from `nuxi` that was using cjs utilities to import/resolve paths within nuxt


src/commands/dev.ts
```diff
@@ -48,7 +48,7 @@ const command = defineCommand({
     // Prepare
     overrideEnv('development')
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir)
-    showVersions(cwd)
+    await showVersions(cwd)
```
pi0, line 51: This was intentionally to block rest of build logic for informational information

PR #640, opened 2025-01-06
perf(module): use native `fetch`

danielroe:
### ❓ Type of change

### 📚 Description

we'll lose the ability to automatically retry fetches, but most functionality of `ofetch` isn't required within `nuxi`

### DIFF
--- a/src/commands/module/_utils.ts
+++ b/src/commands/module/_utils.ts
@@ -1,4 +1,3 @@
-import { $fetch } from 'ofetch'
 import { readPackageJSON } from 'pkg-types'
 import { satisfies, coerce } from 'semver'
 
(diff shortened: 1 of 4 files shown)


src/commands/module/_utils.ts
```diff
@@ -96,9 +95,7 @@ export interface NuxtModule {
 }
 
 export async function fetchModules(): Promise<NuxtModule[]> {
-  const { modules } = await $fetch<NuxtApiModulesResponse>(
-    `https://api.nuxt.com/modules?version=all`,
-  )
+  const { modules } = await fetch('https://api.nuxt.com/modules?version=all').then(r => r.json()) as NuxtApiModulesResponse
```
pi0, line 98: This drops proxy and retry support...

danielroe: can you think of a better approach? this saves us 132kB. I think `retry` isn't really needed, but respecting proxy seems essential to me

pi0: 132kB is %2 (in the next major ofetch/node-fetch-native by dropping Node < 18 the bundle size will be reduced further BTW). 

retry is a good thing (npm/yarn/pnpm has it too). I am hoping in the next iterations of ofetch to implement native fs caching support which contributes to actual runtime performances all this either will be missed for nuxi or has to be rewritten from scratch. Immediately from code above, network error is not handled for example..

Node/undici does not support "system" proxy, it needs a polyfill somewhere which is that bytes.

danielroe: makes sense to me.

handling error would be straightforward, but proxy is key

will continue to look at this

Issue #62, opened 2022-04-04
CLI subcommand support


pi0:
Support `nuxt-` and `nuxi-` external commands same as Nuxt 2. (Current blocker: We need new CLI builder package for shared utils.)

Module builder enhancement: https://github.com/nuxt/module-builder/issues/32

----

References:

- https://github.com/nuxt/rfcs/issues/11
- https://github.com/nuxt/nuxt.js/pull/4314
- Nuxt 2 impl https://github.com/nuxt/nuxt.js/blob/777a4b7f5033c86c37cbd93008f3ca792e4af8bc/packages/cli/src/run.js#L44

PR #644, opened 2025-01-06
feat(subcommand): add subcommand support

danielroe:
### ❓ Type of change

### 📚 Description

resolves https://github.com/nuxt/cli/issues/62

This adds arbitrary CLI subcommand support to `nuxi`. (For example, `@nuxt/module-builder` exposes a `nuxt-module-build` command, which is outdated, as https://github.com/nuxt/module-builder/issues/356 points out.)


src/main.ts
```diff
@@ -1,7 +1,13 @@
+import { resolve } from 'node:path'
+import process from 'node:process'
+
 import { defineCommand } from 'citty'
 import { provider } from 'std-env'
+import { x } from 'tinyexec'
```
pi0, line 6: This import seems could be lazy-loaded to avoid runtime eval overhead unless we hit the fallback

pi0 (approved):
other than comment LGTM.

My main hope in new nuxi with external commands was to avoid executing another process but import citty-compatible modules as command and be able to pass more runtime info but seems far reach today.

danielroe:
oh, that would be nice. I think we can refactor to that as soon as supported. maybe expose a JSON with metadata?

Issue #415, opened 2024-05-20
feat(cleanup): add args to the cleanup command to delete any files


shinGangan:
## 📚 Describe the feature

The cleanup command currently deletes the following directories/files ( https://github.com/nuxt/cli/blob/main/src/utils/nuxt.ts#L18-L30 ) .

- `buildDir`
- `.output`
- `dist`
- `node_modules/.vite`
- `node_modules/.cache

If you want to delete a directory other than the currently specified directory or `node_modules` itself, you have no choice but to rely on the `rimraf` or `rm` command.

I prepare an npm script like below every time.
```json
{
  "scripts": {
    "clean": "rimraf node_modules playground/**/node_modules playground/**/.nuxt",
  }
}
```

Also, in the case of a monorepo configuration using pnpm workspace, `.nuxt`, `node_modules`, and `.output` may not be deleted properly.
Changing `rootDir` option for each directory is a very tedious task.

<br />

Therefore, I would like to add an argument to the `cleanup` command that allows the user to delete arbitrary directories and files.

### Design

An example of usage is as follows.

```sh
npx nuxi@latest cleanup --cleanDir ['.nuxt', 'node_modules']
```

```sh
$ npx nuxi@latest cleanup --help
Clean up generated Nuxt files and caches (nuxi cleanup v3.11.1)                                                          

USAGE nuxi cleanup [OPTIONS] [ROOTDIR]

ARGUMENTS

  ROOTDIR    Root Directory    

OPTIONS

       --cwd    Current working directory
  --logLevel    Log level 
  --cleanDir {any description}
```

shinGangan:
Hi @danielroe , I would appreciate it if you could comment when you have time. 🙏

If there are no problems with the policy, I will start creating PR.

DamianGlowala:
Closing as per the recommendation in https://github.com/nuxt/cli/pull/656#issuecomment-2598180039

PR #656, opened 2025-01-09
feat: additional dirs for cleanDir command

kubicodes:
### ❓ Type of change

### Add custom directory cleanup support to `nuxi cleanup`

This PR adds support for cleaning up custom directories via the `--cleanDir` argument in the `nuxi cleanup` command, as requested in #415.

### Features
- Added `--cleanDir` argument to specify additional directories to clean up
- Supports comma-separated values for cleaning multiple directories in a single command
- Paths are resolved relative to the project root
- Maintains existing functionality for default Nuxt directories

#### Single Directory
```bash
npx nuxi cleanup --cleanDir temp
```

#### Multiple Directories (comma-separated)
```bash
npx nuxi cleanup --cleanDir "temp,cache,build"
```

#### Mixed with Default Cleanup
The command will clean both the specified directories and the default Nuxt directories:
- Custom directories (from --cleanDir)
- buildDir
- .output
- dist
- node_modules/.vite
- node_modules/.cache

### Implementation Details
- Added `cleanDir` argument with `type: 'string'` to support comma-separated values
- Enhanced `cleanupNuxtDirs` to handle custom directories alongside default ones
- All paths are properly resolved relative to the project root
- Whitespace in comma-separated values is automatically trimmed

### DIFF
--- a/src/commands/cleanup.ts
+++ b/src/commands/cleanup.ts
@@ -13,11 +13,21 @@ export default defineCommand({
   args: {
     ...cwdArgs,
     ...legacyRootDirArgs,
+    cleanDir: {
+      type: 'string',
+      required: false,
+      description: 'Additional directories to clean up',
+    },
   },
   async run(ctx) {
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir)
     const { loadNuxtConfig } = await loadKit(cwd)
     const nuxtOptions = await loadNuxtConfig({ cwd, overrides: { dev: true } })
-    await cleanupNuxtDirs(nuxtOptions.rootDir, nuxtOptions.buildDir)
+
+    const customDirs = ctx.args.cleanDir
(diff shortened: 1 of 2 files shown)


danielroe:
don't worry about the failing test - this is probably an issue with how I've set it up

kubicodes:
Hey @danielroe thanks for the heads up :) do you think this change makes sense and could be included?

kubicodes:
> personally I would prefer to implement something like this in a custom module or postinstall script in a project rather than via a flag. is this really needed that often?
> 
> or, given the context of the [original issue](https://github.com/nuxt/cli/issues/415), maybe there's a bug in cleanup or an opportunity to 'do the right thing' automatically?
> 
> what do you think?

I agree with this. I wouldn't say it's needed often and for now I'll just close the PR :) Thanks for the support and input on this 👍

PR #687, opened 2025-01-17
fix(init,module,upgrade): validate response from `consola.prompt`

danielroe:
### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality)
- [ ] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

`consola.prompt` returns a Symbol when `Ctrl-C` is called - we can't trust the types on this one.

PR #741, opened 2025-02-13
fix(nuxi): return true if adding dependency succeed

atinux:
### ❓ Type of change

### 📚 Description

Right now, when running `npx nuxi@latest module add content`, it only install the dependency but does not add it to the `nuxt.config.ts`

I believe the bug has been introduced by this change: https://github.com/nuxt/cli/pull/687/files#diff-3ac0a0d31e5ed2e2949d42ea2e33d33bf48d9663ae4031a23a2582c31cc4f2af

`nypm` does not return `true` when installing a dependency: https://github.com/unjs/nypm/blob/651ab80c383bd74dde42de06763240e14611473c/src/api.ts#L59C23-L59C36

My changes make sure that we return `true` if `nypm` did not throw any error so it can move to updating the `nuxt.config` file

### DIFF
--- a/packages/nuxi/src/commands/module/add.ts
+++ b/packages/nuxi/src/commands/module/add.ts
@@ -132,18 +132,20 @@ async function addModules(modules: ResolvedModule[], { skipInstall, skipConfig, 
         cwd,
         dev: isDev,
         installPeerDependencies: true,
-      }).catch(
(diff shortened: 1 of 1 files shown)


atinux:
Yes it is! Nice catch @BobbieGoede

PR #742, opened 2025-02-13
feat(init): allow installing modules after clone

### LINKED ISSUE #267: Adding modules when creating the project
### Describe the feature

I was thinking there should be a way to add modules to the project when creating it. Heres an example:

`npx nuxi@latest init project-name -M tailwindcss eslint`

The -M denotes the modules to be installed and added to the Nuxt config.

### Additional information

- [ ] Would you be willing to help implement this feature?
- [ ] Could this feature be implemented as a module?

### Final checks

- [X] Read the [contribution guide](https://nuxt.com/docs/community/contribution).
- [X] Check existing [discussions](https://github.com/nuxt/nuxt/discussions) and [issues](https://github.com/nuxt/nuxt/issues).

kubicodes:
### 🔗 Linked issue
resolves https://github.com/nuxt/cli/issues/267

### ❓ Type of change

### 📚 Description

An additional flag `-M` was added for `nuxi init`  .
It accepts a string of nuxt modules, as an example : `nuxi init -M eslint,tailwind`

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -10,6 +10,7 @@ import { installDependencies } from 'nypm'
 import { relative, resolve } from 'pathe'
 import { x } from 'tinyexec'
 
+import { runCommand } from '../run'
 import { logger } from '../utils/logger'
 import { cwdArgs } from './_shared'
 
@@ -73,6 +74,12 @@ export default defineCommand({
     packageManager: {
       type: 'string',
       description: 'Package manager choice (npm, pnpm, yarn, bun)',
+    },
+    modules: {
+      type: 'string',
+      required: false,
+      description: 'Modules to install',
+      alias: 'M',
     },
   },
   async run(ctx) {
@@ -203,6 +210,12 @@ export default defineCommand({
       catch (err) {
         logger.warn(`Failed to initialize git repository: ${err}`)
       }
+    }
+
+    // Add modules when -M flag is provided
+    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(' ').filter(name => !!name)
+    if (modules.length > 0) {
+      await runCommand('module', ['add', ...modules])
     }
 
     // Display next steps
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -213,7 +213,7 @@ export default defineCommand({
     }
 
     // Add modules when -M flag is provided
-    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').filter(module => !!module.trim())
+    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').map(module => module.trim()).filter(Boolean)
     if (modules.length > 0) {
       await runCommand('module', ['add', ...modules])
     }
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -78,7 +78,7 @@ export default defineCommand({
     modules: {
       type: 'string',
       required: false,
-      description: 'Modules to install',
+      description: 'Nuxt modules to install (comma separated without spaces)',
       alias: 'M',
     },
   },
@@ -213,7 +213,7 @@ export default defineCommand({
     }
 
     // Add modules when -M flag is provided
-    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(' ').filter(name => !!name)
+    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').map(module => module.trim())
     if (modules.length > 0) {
       await runCommand('module', ['add', ...modules])
     }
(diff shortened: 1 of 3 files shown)


packages/nuxi/src/commands/init.ts
```diff
@@ -74,6 +75,12 @@ export default defineCommand({
       type: 'string',
       description: 'Package manager choice (npm, pnpm, yarn, bun)',
     },
+    modules: {
+      type: 'string',
```
kubicodes, line 79: this requires to provide multiple modules in quotes, such as "esllint tailwind". not sure about the user experience here. when providing it as `-M tailwind eslint` it just reads `tailwind` on `ctx.modules` . Does anyone know if it would be possible without the quotes?

danielroe: let's go for a string of modules separated by commas, which will work even without quotes

danielroe:
This looks great - next time you create a PR, would you mind ticking the box to give maintainers the ability to push changes to the PR? 🙏

kubicodes:
> This looks great - next time you create a PR, would you mind ticking the box to give maintainers the ability to push changes to the PR? 🙏

Hey, thanks for the review :) Sure, I think i accidentally removed the tick. If it's necessary I can create a new PR, just lmk 🙏

danielroe:
No worries! I was just going to push:

```ts
const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').filter(name => !!name.trim())
```

packages/nuxi/src/commands/init.ts
kubicodes: I tried this right now with -M tailwind,eslint and ctx.args.modules is then defined as tailwind, . Am I maybe missing something? 🤔

danielroe: CleanShot 2025-02-13 at 21 51 12@2x

kubicodes: Ah I see the issue now. I tried with spaces to cover the edge case, so my input were:
`-M eslint, tailwind` and `-M eslint,    tailwind` . Whenever there's a space after the comma it automatically trims it and it's not available in the `ctx.args.modules` anymore. 
I just pushed the change for the `trim()` which assumes a comma separated list without spaces. also adjusted the description of the flag to mention that it assumes a comma separated list without spaces.

should we investigate that issue further or is it fine for now?

packages/nuxi/src/commands/init.ts
```diff
@@ -213,7 +213,7 @@ export default defineCommand({
     }
 
     // Add modules when -M flag is provided
-    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(' ').filter(name => !!name)
+    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').map(module => module.trim())
```
danielroe, line 216: I think we still need !! because if the user quotes the list, as in `--modules 'test, thing'` they will need to be trimmed

packages/nuxi/src/commands/init.ts
danielroe: It's fine - spaces normally indicate an 'end' of the argument, when used in a CLI context, so this is what I would expect.

packages/nuxi/src/commands/init.ts
kubicodes: I think it would work, because now I'm using a `map` rather than a `filter`. So for example if we have

`"one,     two,     threee"` as input, the `split()` would return `[ 'test', '     thing', '     third' ]` and each of them gets trimmed during the `map`. 

Screenshot 2025-02-13 at 22 09 18

Am I missing something?

danielroe: no, you're quite right - I should have given the example of `--modules test,`

kubicodes: ah amazing, great catch! I removed the `map` now according to https://github.com/nuxt/cli/pull/742#discussion_r1955213279 
and added the `trim` function according to: https://github.com/nuxt/cli/pull/742#issuecomment-2657663792

packages/nuxi/src/commands/init.ts
```diff
@@ -213,7 +213,7 @@ export default defineCommand({
     }
 
     // Add modules when -M flag is provided
-    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').map(module => module.trim())
+    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').filter(module => !!module.trim())
```
danielroe, line 216: you still need to map, I think, as well as filter

danielroe:
```suggestion
    const modules = !ctx.args.modules ? [] : ctx.args.modules.split(',').map(module => module.trim()).filter(Boolean)
```

kubicodes: good point, I just removed it because we don't expect any spaces between the comma separated values. however we'd still need to cover this case `"   module1,module2` to trim the first value. I commited the suggestion :)

danielroe (approved):
❤️

Issue #697, opened 2025-01-22
Add option to specify version when using `npx nuxi@latest module add`


maximepvrt:
The `npx nuxi@latest module add` command does not currently allow users to specify the version of the module to install.  

I discovered this limitation while trying to update the documentation for Nuxt Content v2. Since the release of Nuxt Content v3 last week, the `npx nuxi@latest module add content` command now installs v3 by default, which is not suitable for projects still using v2 [https://v2.content.nuxt.com/get-started/installation#add-to-a-project](https://v2.content.nuxt.com/get-started/installation#add-to-a-project)

schirrel:
@maximepvrt to install a specific version of a module you will need to use the whole org/plg/version syntax:
- `npx nuxi@latest module add @nuxt/content@2.13.0`

![Image](https://github.com/user-attachments/assets/826121e0-6941-44e5-8388-abcf75630529)

maximepvrt:
@danielroe @schirrel However, it still doesn't work if the full version isn't specified. The commands `npx nuxi@latest module add content@2` and `npx nuxi@latest module add @nuxt/content@2` doesn't run properly.

maximepvrt:
@danielroe can you reopen ?

schirrel:
@maximepvrt @danielroe hang on i think i already know how to fix. Will try to open the PR in around 6 hours

schirrel:
@maximepvrt @danielroe can you guys test the code?
https://github.com/nuxt/cli/pull/803

PR #803, opened 2025-03-19
fix(module): resolve semver module versions

schirrel:
### 🔗 Linked issue

Fix for https://github.com/nuxt/cli/issues/697
### ❓ Type of change

### 📚 Description

The code verify if the `pkgVersion` is only a single number to then perform a search for the highest major version.

Previously PR fix part of the problem but not allow to install full package from a single major version.
This PR enhance the `nuxi module add` to allowing it to find a Nuxt Module full version when only a major is specified.
Without this PR:
Screenshot 2025-03-19 at 17 20 44

With this PR:

Screenshot 2025-03-20 at 06 53 52

Regression:
> default installation not affected
Screenshot 2025-03-19 at 17 19 53

-----------------------------------------------------------------------

### DIFF
--- a/packages/nuxi/src/commands/module/add.ts
+++ b/packages/nuxi/src/commands/module/add.ts
@@ -289,7 +289,7 @@ async function resolveModule(moduleName: string, cwd: string): Promise<ModuleRes
   }
 
   // Fetch package on npm
-  pkgVersion = pkgVersion || 'latest'
+  let version = pkgVersion || 'latest'
   const pkgScope = pkgName.startsWith('@') ? pkgName.split('/')[0]! : null
   const meta: RegistryMeta = await detectNpmRegistry(pkgScope)
   const headers: HeadersInit = {}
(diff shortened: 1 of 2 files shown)


packages/nuxi/src/commands/module/add.ts
```diff
@@ -305,6 +305,11 @@ async function resolveModule(moduleName: string, cwd: string): Promise<ModuleRes
   // check if a dist-tag exists
   pkgVersion = (pkgDetails['dist-tags']?.[pkgVersion] || pkgVersion) as string
 
+  // if a the major version was provided, search for the latest version that starts with that number
+  if (pkgVersion && pkgVersion.length === 1) {
+    pkgVersion = Object.keys(pkgDetails.versions)?.findLast(version => version.startsWith(pkgVersion as string)) || pkgVersion
```
danielroe, line 310: can we use `semver`'s `satisfies` so we support any version range?

maximepvrt:
Awesome, I just tested it, but it doesn't work with minor versions:  `nuxi module add @nuxt/content@2.13`

packages/nuxi/src/commands/module/add.ts
schirrel: For sure Dan, working on it

schirrel: Fixed [c734d48](https://github.com/nuxt/cli/pull/803/commits/c734d48482935e3de64cc3e603c1a311344fabf9)

schirrel:
@maximepvrt @danielroe all set and done guys 🙏🏻

packages/nuxi/src/commands/module/add.ts
```diff
@@ -305,6 +305,11 @@ async function resolveModule(moduleName: string, cwd: string): Promise<ModuleRes
   // check if a dist-tag exists
   pkgVersion = (pkgDetails['dist-tags']?.[pkgVersion] || pkgVersion) as string
 
+  // if not a full x.y.z version is provided, search for the latest version that starts with that version
+  if (pkgVersion && pkgVersion.split('.').length < 3) {
+    pkgVersion = Object.keys(pkgDetails.versions)?.findLast(version => version.startsWith(pkgVersion as string)) || pkgVersion
```
danielroe, line 310: would you use `satisfies` from `semver? 🙏

danielroe (approved):
this looks good to me! let me know if I've missed something in your implementation when porting it to use `semver` ranges 🙏

schirrel:
@danielroe sure, i was making some experiments on using semver lib but could not finish that, as soon as i can i will update that.
Thanks

danielroe:
no worries! i pushed directly

PR #823, opened 2025-04-01
fix(preview): load nuxt instance to get final nitro output path

### LINKED ISSUE #804: `nuxi preview/start` (using kit's `loadNuxtConfig`) not resolving installed nuxt modules, ends up with partial config
Hello,

We currently have an issue running nuxi's preview/start due to how we setup the config, as preview uses `loadNuxtConfig` from `@nuxt/kit` instead of `loadNuxt`. Let me explain the use case:

We got a number of Vue apps which we created a base config that all of them follow _(includes server things, Vue defaults as router and vuex, test configuration, app configuration, and so on)_. 

When planning to move them to Nuxt, our idea was to create a Nuxt Module project which will be installed in every project, and it'll provide our base configurations for all apps.

## setup 

"cool-nuxt-config" project:

```
    export default defineNuxtModule({
        async setup(_options, _nuxt) {
            const ourCoolDistFolder = 'dist/nested/folder/dont/ask/me/why';

            // various nuxt.config.ts "base" configurations here
            // such as nitro, vite, app.head, etc
            const baseConfig = {
                nitro: {
                    output: {
                        dir: ourCoolDistFolder,
                        serverDir: `${ourCoolDistFolder}/server`,
                        publicDir: `${ourCoolDistFolder}/public`,
                    },
                },
                (..etcetc..)
            };

            // more code like installModule for nuxt-primevue, for e.g., and so on
            (..etcetc..)


            // using defu, we make sure our config has priority when merging app's own configs
            _nuxt.options = defu(baseConfig, _nuxt.options)
        }
    });
```

And later, we install it as a module in the App `nuxt.config.ts`:

```
        nitro: {
            // fake test prop explained below
            testexample: 'value'
        },

        modules: [
            'cool-nuxt-config'
        ],
    });
```

When running `nuxi build`, it's all good. Build uses kit's `loadNuxt` ( [source](https://github.com/nuxt/cli/blob/main/packages/nuxi/src/commands/build.ts#L44) ), which also resolves modules too, so that `nuxt.options` will contain our nitro options, that is, `nuxt.options.nitro.output.dir` will reflect our `ourCoolDistFolder` and will create the built project into the path we chose instead of `.output`.

## The issue

Different than the build, preview on the other hand uses kit's `loadNuxtConfig` ( [source](https://github.com/nuxt/cli/blob/main/packages/nuxi/src/commands/preview.ts#L36) ) which does **NOT** evaluate/run modules installed. That means our module's setup is never called, which means `config.nitro.testexample` _(from root's nuxt.config.ts, as I showed on code excerpt above)_ is defined, but `config.nitro.output.dir` is not defined since modules were not run. that entails in [resolvedOutputDir](https://github.com/nuxt/cli/blob/main/packages/nuxi/src/commands/preview.ts#L42-L49) ALWAYS resolving into `./.output/nitro.json` (excerpt below):

```
    const resolvedOutputDir = resolve(
      config.srcDir || cwd,
      config.nitro.output?.dir || '.output',
      'nitro.json',
    )
```

I tried giving `cwd` our path just to test, but doesn't matter because the second argument in `resolve` is `config.nitro.output?.dir || '.output',` and is always resolving into `.output` since `config.nitro.output.dir` is not defined. And this means preview never works, because it always dies on the message: 

> "Cannot find nitro.json. Did you run nuxi build first?"

I did a test and changing

```
    const { loadNuxtConfig } = await loadKit(cwd)
    const config = await loadNuxtConfig({
      cwd,
      envName: ctx.args.envName, // c12 will fall back to NODE_ENV
      overrides: /* ctx.options?.overrides || */ {},
    })
```

to

```
    const { loadNuxt } = await loadKit(cwd)
    const nuxt = await loadNuxt({
      cwd,
      envName: ctx.args.envName, // c12 will fall back to NODE_ENV
      overrides: /* ctx.options?.overrides || */ {},
    })
    const config = nuxt.options
```

makes it so preview can now properly find `nitro.json` using the module's `nitro.output.dir` configuration.

## questions

- Is there any reason why preview uses loadNuxtConfig instead of loadNuxt? That prevents Nuxt Modules and Layers configs from being applied, which, in this user case I presented, brings nitro and other configurations.
- Is it okay to open a PR with the change I mentioned above to `loadNuxt` instead? 
- Do you see any other way to keep shared configuration across multiple Nuxt projects other than the one we thought, using Nuxt Module instead? _(though we thought that's the point of Module :D )_

Thank you

RaphaelDDL:
### 🔗 Linked issue

resolves https://github.com/nuxt/cli/issues/804

### ❓ Type of change

### 📚 Description

As stated in https://github.com/nuxt/cli/issues/804 , different than build and other commands (analyze, prepare, etc), the `preview` command (and the alias `start`) uses nuxt/kit `loadNuxtConfig` instead of `loadNuxt`. That entails in reading only the app nuxt.config.ts, but not resolving any modules. That means that modules which contain configuration such as the nitro output directories will not be evaluated, meaning preview will try to resolve the output path only with whatever configuration is in root config.

### Testing

- do the steps as per "contributing" in readme
- In `packages/nuxt-cli/playground` _(referred as "playground" from now on)_, run `pnpm nuxi build`
- Verify that `.nuxt` and `.output` were created as per default configuration,  run `pnpm nuxi preview` to verify server starts
- delete the two folders (important otherwise `.output` will be wrongly used by the scenario below)
- create `module.ts` in `packages/nuxt-cli/playground/module.ts` with the following  _(more info on usage in the issue)_:

```
  import { defineNuxtModule } from '@nuxt/kit';
  import { defu } from 'defu';
  
  const outputPath = `dist/a/b/c`;
  
  export default defineNuxtModule({
    async setup(_options, _nuxt) {
      const baseConfig = {
        nitro: {
          compressPublicAssets: true,
          output: {
            dir: outputPath,
            serverDir: `${outputPath}/server`,
            publicDir: `${outputPath}/public`,
          },
        },
      };
      _nuxt.options = defu(baseConfig, _nuxt.options);
    },
  });
```
- `pnpm add @nuxt/kit` in playground (`packages/nuxt-cli/playground`) in order to use `defineNuxtModule`.
- Add the following in `packages/nuxt-cli/playground/nuxt.config.ts`
```
  modules: [
    './module.ts'
  ]
```
- run `pnpm nuxi build`
- verify that this time it was created `.nuxt` and `dist` (message will be `Generated public dist/a/b/c/public` and `You can preview this build using node dist/a/b/c/server/index.mjs`)
- run `pnpm nuxi preview`
- You will get an error stating `Cannot find nitro.json. Did you run nuxi build first?` due to `loadNuxtConfig` not evaluating module's configs
- Apply this PR change
- run `pnpm nuxi preview`
- Server started as found nitro.json in the `nitro.output.dir` as configured by the Nuxt Module `dist/a/b/c/nitro.json`

As Daniel pointed out in the issue, it will indeed slow down a bit the preview start, in playground took less than 2s with current preview and 3s with loadNuxt.

Please let me know if there's anything I must change.
Thank you for your time checking the issue.


danielroe:
thank you! ❤️

PR #879, opened 2025-05-13
fix(init): skip modules that are dependencies of other modules

### LINKED ISSUE #875: Never install `@nuxt/fonts` or `@nuxt/icon` if `@nuxt/ui` is selected
In the wizard for init a project, if the user ticks one of the official modules that is already installed by default by @nuxt/ui, the module should not be installed because it is not useful in the first place and in some cases it may bug the configurations applied by default by nuxt/ui.

https://github.com/nuxt/cli/blob/main/packages/nuxi/src/commands/init.ts#L270

HugoRCD:
### 🔗 Linked issue

Resolves #875 

### ❓ Type of change


packages/nuxi/src/commands/init.ts
```diff
@@ -297,7 +334,19 @@ export default defineCommand({
         }
 
         if (selectedOfficialModules.length > 0) {
-          modulesToAdd.push(...(selectedOfficialModules as unknown as string[]))
+          const modules = selectedOfficialModules as unknown as string[]
+
+          const allDependencies: Record<string, string[]> = {}
+          for (const module of modules) {
+            allDependencies[module] = await getModuleDependencies(module)
```
danielroe: we probably also want to get the dependencies of the template that's been installed (and not prompt to install any modules it contains), but that can be for another pr

for example, if you run `pnpm create nuxt -t ui`, it probably should not prompt to install `@nuxt/fonts`.

danielroe (approved):

packages/nuxi/src/commands/init.ts
HugoRCD: I'll try to make a PR for that too!

Issue #887, opened 2025-05-24
Modules prompt triggers when empty module list is passed


cjpearson:
When the modules flag is used for init, the prompt still shows if the list of modules is empty.

**Current behavior**

`npx init` => prompt shown ✅ 
`npx init --modules '@nuxt/fonts'` => no prompt ✅ 
`npx init --modules` => prompt shown ❌ 

**Expected**

`npx init` => prompt shown ✅ 
`npx init --modules '@nuxt/fonts'` => no prompt ✅ 
`npx init --modules ''` => no prompt ✅

PR #888, opened 2025-05-24
fix: do not prompt for modules if an empty list is provided

cjpearson:
### 🔗 Linked issue

Resolves https://github.com/nuxt/cli/issues/887

### ❓ Type of change

### 📚 Description

When `--modules` is not used, `ctx.args.modules` will be `undefined` and when `--modules ''`  or `--modules` is used, it will be an empty string. With this change, we will only prompt in the `undefined` case, allowing users to explicitly skip the prompt if they want.

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -85,6 +85,7 @@ export default defineCommand({
       type: 'string',
       required: false,
       description: 'Nuxt modules to install (comma separated without spaces)',
+      negativeDescription: 'Skip module installation prompt',
       alias: 'M',
     },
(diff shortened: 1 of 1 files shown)


danielroe:
maybe we could support and document '--no-modules'

cjpearson:
That could definitely work too. I think it's more direct, but I wasn't sure how to handle if it's combined with `--modules`. If `--no-modules` is passed should it only skip the prompt or also skip the provided modules?

danielroe:
both i think. citty handles --no prefix for all args i think ...

danielroe (approved):

PR #933, opened 2025-07-01
feat: add `--extends` option to support nuxt layer extension

### LINKED ISSUE #930: allow --extends to extend a nuxt layer
https://github.com/nuxtlabs/docus/issues/1096

ideally, we'd like to avoid libraries or projects creating their own clis over nuxt/cli, as this opts them out of any performance improvements that we merge into nuxt/cli, and also diverges the ecosystem.

a new `--extends` cli option would enable `docus build` (for example) to be replaced with `nuxt build --extends docus`

sanjayy-gowdaa:
### 🔗 Linked issue
Resolves #930 

### ❓ Type of change

### 📚 Description

This PR adds a new --extends CLI option to core Nuxt commands, enabling users to extend Nuxt layers directly from the command line. This addresses the need for a unified CLI experience across the Nuxt ecosystem.

Problem: Currently, projects like Docus need to create their own CLI tools (e.g., docus build)

Changes:

- Add extendsArgs to _shared.ts for consistent argument definition
- Add --extends support to core commands: build, dev, generate, preview
- Add --extends support to utility commands: analyze, typecheck, prepare


packages/nuxi/src/commands/dev.ts
```diff
@@ -100,6 +101,7 @@ const command = defineCommand({
       overrides: {
         dev: true,
         logLevel: ctx.args.logLevel as 'silent' | 'info' | 'verbose',
+        ...(ctx.args.extends && { extends: ctx.args.extends }),
```
danielroe, line 104: this also needs to be passed lower down as well when starting the dev server.

sanjayy-gowdaa: I've now updated the implementation to pass the extends argument down to the dev server level as well.

danielroe (approved):
this looks great - thank you ❤

PR #957, opened 2025-07-14
fix(dev): temporarily refactor to use previous socket url handling

danielroe:
This reverts commit c7c4dd100ed0488bc84f97d3610b0d30474f9e73.

### 📚 Description

this temporarily reverts https://github.com/nuxt/cli/pull/952 - we'll identify the issue later if this solves it, but I'm temporarily without access to windows


OrbisK:
> I'm temporarily without access to windows

I think you dont need windows. This does even reproduce in github codespaces.

PR #998, opened 2025-08-08
fix(init): use amended directory when adding modules

AaronDewes:
### 🔗 Linked issue

I could not find one, let me know if I should open one.

### ❓ Type of change

### 📚 Description

Previously, when the user selected a directory that already existed, and was then prompted to choose a different directory, and then selected a different directory, and also chose official modules to add, the official modules would be added to the initial directory, not the newly chosen one.


danielroe (approved):
good find!

Issue #484, opened 2024-09-25
feat(upgrade): implement `to` argument for precise version control


OrbisK:
This feature allows precise control over the desired version for Nuxt and all its sub-packages. It provides the flexibility to, for instance, update only to the next minor version.

~~The more I look at it, the bigger the changes. I've just seen that the upgrade effectively only deletes the lockfile and updates everything. Is there no way to just update all nuxt deps?~~

~~Cant nuxi upgrade install nuxt version and dedupe lockfile with the same effect?~~

EDIT: dedupe is now a thing 😄

OrbisK:
this will allow renovate https://docs.renovatebot.com/self-hosted-configuration/#allowpostupgradecommandtemplating precise post update scripts

dervondenbergen:
I just wanted to open a new issue and saw, that his one already exists.

My usecase is sadly not solved by `--dedupe` as I only want to upgrade from a `v3.16.3` to a `v3.17.7`, so only minor + patch. When I run the upgrade command now I get `v4.0.0` where things are breaking altough I only want to update to the latest minor and see if everything works there before upgrading major version.

Would this be the same issue or should I open a new ticket for my usecase?

danielroe:
`channel` feels good to me....

PR #1019, opened 2025-08-26
feat(upgrade): add `v3`, `v4`, `v3-nightly` and `v4-nightly` channel

OrbisK:
### 🔗 Linked issue

resolves https://github.com/nuxt/cli/issues/484

### ❓ Type of change

### 📚 Description

This PR introduces the following new channels: `v3`, `v4`, `v3-nightly` and `v4-nightly`.

This should allow `nuxt upgrade` to be used to upgrade minor versions of v3
```sh
pnpx nuxt@latest upgrade --dedupe --channel=v3
```

### DIFF
--- a/packages/nuxi/src/commands/upgrade.ts
+++ b/packages/nuxi/src/commands/upgrade.ts
@@ -62,7 +62,7 @@ async function getRequiredNewVersion(packageNames: string[], channel: string): P
     case 'v3-nightly':
       return { npmPackages: packageNames.map(p => getNightlyDependency(p, '3.x')), nuxtVersion: '3.x' }
     case 'v4':
-      return { npmPackages: packageNames.map(p => `${p}@latest`), nuxtVersion: '4.x' }
+      return { npmPackages: packageNames.map(p => `${p}@4`), nuxtVersion: '4.x' }
     case 'v4-nightly':
       return { npmPackages: packageNames.map(p => getNightlyDependency(p, '4.x')), nuxtVersion: '4.x' }
     case 'stable':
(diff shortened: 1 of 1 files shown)


danielroe (commented):
this looks good to me! only question is about `4.x` vs `4x` vs `v4` vs `4` as channel name

also, thinking about forward compatibility - I think we should maybe start releasing v4 as 4x _and_ latest so we do not need to hard-code `latest` in here for `v4` channel

(creating PR to do so in Nuxt)

packages/nuxi/src/commands/upgrade.ts
```diff
@@ -42,17 +48,27 @@ async function getNightlyVersion(packageNames: string[]): Promise<{ npmPackages:
...
+async function getRequiredNewVersion(packageNames: string[], channel: string): Promise<{ npmPackages: string[], nuxtVersion: NuxtVersionTag }> {
+  switch (channel) {
+    case 'nightly':
+      return getNightlyVersion(packageNames)
+    case 'v3':
+      return { npmPackages: packageNames.map(p => `${p}@3`), nuxtVersion: '3.x' }
+    case 'v3-nightly':
+      return { npmPackages: packageNames.map(p => getNightlyDependency(p, '3.x')), nuxtVersion: '3.x' }
+    case 'v4':
+      return { npmPackages: packageNames.map(p => `${p}@latest`), nuxtVersion: '4.x' }
```
OrbisK, line 65:
```suggestion
      return { npmPackages: packageNames.map(p => `${p}@4`), nuxtVersion: '4.x' }
```

PR #1021, opened 2025-08-31
feat(build,dev): add vite and vue version to banner

OskarLebuda:
### 🔗 Linked issue

Resolves https://github.com/nuxt/nuxt/issues/33071

### ❓ Type of change

### 📚 Description

Add version of `vite` (also `rolldown-vite`) and `vue` to the cli banner


packages/nuxi/src/commands/dev.ts
```diff
@@ -92,7 +92,7 @@ const command = defineCommand({
     // Prepare
     overrideEnv('development')
     const cwd = resolve(ctx.args.cwd || ctx.args.rootDir)
-    showVersions(cwd)
+    await showVersions(cwd)
```
TheAlexLichter, line 95: Do we need to await it here? I wonder if it matters order-wise and would improve dev server startup time / build speed (for `build.ts`) if we would not await it (so it can run in parallel)

OskarLebuda: Yes, you're right 🙂 removed

TheAlexLichter (approved):

PR #1043, opened 2025-09-17
fix(init): properly check installed modules

atinux:
- The detection of the installed modules was not working, using `await import()` was wrong, fixed it by using `readPackageJSON`
- I also detected ahead of time if nuxt is installed as dependency to avoid asking to install Nuxt modules
- This gives the possibility to install the Nuxt UI templates with Vite while keeping our nice Nuxt branding and CLI

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -77,7 +77,7 @@ async function getTemplateDependencies(templateDir: string) {
     if (!existsSync(packageJsonPath)) {
       return []
     }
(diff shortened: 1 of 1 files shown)


atinux:
This is what I thought but wanted to confirm with you.

It's just that the DX of nuxt init is sexy (confirm to overwrite if dir exists, ask to install deps, etc)

pi0:
> It's just that the DX of nuxt init is sexy (confirm to overwrite if dir exists, ask to install deps, etc)

We could add this features to giget CLI too btw.

danielroe (approved):

PR #1050, opened 2025-09-24
fix: Triggering a loop restart issue when saving the nuxt.config.ts file

jianxing-xu:
### 🔗 Linked issue

refer: 
https://github.com/nuxt/bridge/issues/1591
https://github.com/nuxt/bridge/issues/1623

### ❓ Type of change

### 📚 Description

I'm migrating my Nuxt2 app to Bridge and I'm running into two issues.

First, I received a response in issue 1591, stating that this issue is caused by versions later than nuxi@3.6.0. So I rolled back to nuxi@3.6.0, which then caused another issue: https://github.com/nuxt/bridge/issues/1623 The app is only reloaded when nuxt.config.js is first saved. I found the answer in the source code: during onReload , the close method is not executed, so the watchConfig method is only executed during the initial startup and init .

Second issue: After using the latest version of Nuxi, I encountered the issue https://github.com/nuxt/bridge/issues/1591. After saving the nuxt.config.js file, the dev server would restart infinitely. I eventually discovered the cause: it was due to the different build environments for Nuxt 2 and Nuxt 3. In Nuxt 2's "kit.buildNuxt" step in the "load" method, Nuxt 2 first cleared the .nuxt directory and then regenerated the artifacts, but Nuxt 3+ didn't do this. As a result, modifications to the dist directory in Nuxt 2 triggered the watcher "this._distWatcher = watch(distDir)", causing the app to restart and then begin a loop. The fix is ​​actually quite simple: simply execute the "close" method before the "_load" method and then re-execute the "watchConfig" method afterward.

Finally, this is a fix specifically for the bridge environment; I don't know if it will make it into the main version of Nuxi.
image
image
image
image

### DIFF
--- a/packages/nuxi/src/dev/utils.ts
+++ b/packages/nuxi/src/dev/utils.ts
@@ -160,7 +160,9 @@ export class NuxtDevServer extends EventEmitter<DevServerEventMap> {
 
   async load(reload?: boolean, reason?: string) {
     try {
+      this.closeWatchers();
       await this._load(reload, reason)
+      this._watchConfig();
       this._loadingError = undefined
     }
     catch (error) {


danielroe:
thank you for the PR & investigation 🙏

PR #1137, opened 2025-11-21
fix(dev): prime mtimes cache when starting dev server

### LINKED ISSUE #1135: Dev server constantly restarting on Windows in v3.30.0
The problem described in #1028 , introduced in v3.26.0 by #913, and fixed in v3.29.0 by #1060, has returned in v3.30.0.

I believe the regression was introduced by #1105, which introduced the new `ForkPool`. The issue appears to be that the `FileChangeTracker`, which was introduced to address the problem originally, is re-created with each new fork. This class has an `mtimes` cache of previously-known file update times, and so must persist across restarts to be useful. (Or, perhaps that cache just needs to be populated at the time the `FileChangeTracker` is created? I'm a bit unclear how it was working previously.)

Tagging @danielroe, as the author of all the mentioned changes.

aroneous:
### 🔗 Linked issue

Resolves #1135

### ❓ Type of change

### 📚 Description

When watching for file/directory changes, prime the FileChangeTracker with initial mtimes of the tracked files, such that if the first notification received for a file is spurious (does not represent an actual change), shouldEmitChange will return 'false'.

Without this, the 'priming' happened via several reloads/restarts, but with the introduction of the ForkPool, the FileChangeTracker cache no longer perists across restarts, never allowing it to become populated. Result: unpredictable restarts.

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -21,6 +21,8 @@ jobs:
 
       - name: 📦 Install dependencies
         run: pnpm install
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: 🔠 Lint project (+ fix)
         run: pnpm run lint:fix
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -26,6 +26,8 @@ jobs:
 
       - name: 📦 Install dependencies
         run: pnpm install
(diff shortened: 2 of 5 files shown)


aroneous:
Fixed Windows-specific issues (yeah, the fix is for Windows in the first place, but I developed in on a Mac for silly reasons). I don't think the remaining test failure is related to this change.

PR #1169, opened 2025-12-23
fix(dev): preserve user-defined HMR port in vite:extend hook

productdevbook:
## Summary

Remove `port: undefined` override that was clobbering user-specified HMR port, causing port conflicts in monorepo setups.

## Changes

```diff
config.server.hmr = {
  protocol: undefined,
  ...(config.server.hmr as Exclude<typeof config.server.hmr, boolean>),
- port: undefined,
  host: undefined,
  server: this.listener.server,
}
```

The spread operator already preserves the port from user config, so explicitly setting `port: undefined` was unnecessary and broke custom port configuration.

## Related

- nuxt/cli#1170
- nuxt/nuxt#33929

### DIFF
--- a/packages/nuxi/src/dev/utils.ts
+++ b/packages/nuxi/src/dev/utils.ts
@@ -343,7 +343,6 @@ export class NuxtDevServer extends EventEmitter<DevServerEventMap> {
           config.server.hmr = {
             protocol: undefined,
             ...(config.server.hmr as Exclude<typeof config.server.hmr, boolean>),
-            port: undefined,
             host: undefined,
             server: this.listener.server,
           }


packages/nuxi/src/dev/utils.ts
```diff
@@ -340,10 +340,12 @@ export class NuxtDevServer extends EventEmitter<DevServerEventMap> {
     if (!process.env.NUXI_DISABLE_VITE_HMR) {
       this.#currentNuxt.hooks.hook('vite:extend', ({ config }) => {
         if (config.server) {
+          // Preserve user-defined HMR port to prevent conflicts in monorepo setup
+          const existingPort = typeof config.server.hmr === 'object' ? config.server.hmr.port : undefined
           config.server.hmr = {
             protocol: undefined,
             ...(config.server.hmr as Exclude<typeof config.server.hmr, boolean>),
-            port: undefined,
+            port: existingPort,
```
danielroe: wouldn't it have the same effect simply to delete this line?

productdevbook: Maybe, Right now my head is completely confused, I'm rethinking everything. I hope I can get it working one last time.

productdevbook: Yes thats true

PR #1235, opened 2026-02-27
fix(nuxi): keep cwd at project root during preview

Davidson3556:
### 🔗 Linked issue

resolves nuxt/cli#1368

### 📚 Description

`nuxt preview` was spawning the server with `cwd` set to the `.output` directory, 
which means `process.cwd()` inside server handlers returns `.output` instead of 
the project root. This is inconsistent with running `node .output/server/index.mjs` 
directly, which keeps the cwd at project root.

The fix keeps `cwd` as the project root when spawning the preview server, and 
resolves the server entrypoint args to absolute paths using `outputPath` so they 
work without needing a chdir.

### DIFF
--- a/packages/nuxi/src/commands/preview.ts
+++ b/packages/nuxi/src/commands/preview.ts
@@ -1,23 +1,30 @@
-import { existsSync, promises as fsp } from 'node:fs'
-import { dirname } from 'node:path'
-import process from 'node:process'
+import { existsSync, promises as fsp } from "node:fs";
+import { dirname } from "node:path";
+import process from "node:process";
 
-import { box, outro } from '@clack/prompts'
-import { setupDotenv } from 'c12'
-import { defineCommand } from 'citty'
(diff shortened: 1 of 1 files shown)


Davidson3556:
The runtimes.spec.ts failures on Windows are unrelated to this change — they're caused by a port conflict on the CI runner (port 3100 timeout). The same test suite passes for bun and deno runtimes, and the preview test in commands.spec.ts passes cleanly.

Davidson3556:
check successful

danielroe (commented):
`process.cwd()` is probably not something you should be using within your nitro endpoints. it's very tied to node and isn't portable. why are you doing it?

if it's to access files on disk, I think Nitro's storage layer is a better fit for what you want.

danielroe:
I would love to improve the displayed command. this would need to happen upstream in nitro I think 🙏

PR #1264, opened 2026-03-26
feat(init): detect agent/CI environments and skip interactive prompts

atinux:
Uses `std-env` v4's `isAgent`, `isCI`, and `hasTTY` to automatically
detect when the CLI is run non-interactively (AI agents, CI pipelines,
piped input) and apply sensible defaults instead of hanging on prompts.

Adds two explicit flags for opt-in:
- `--defaults` / `-y`: accept all defaults (like `npm init -y`)
- `--no-interactive`: same, more explicit for scripting contexts

When non-interactive mode is detected, the CLI:
- Logs the reason (agent name, CI, no TTY, or flag)
- Displays a full options reference — including all available templates
  fetched live — so agents can discover flags and re-run with custom
  settings
- Auto-selects: template=minimal, dir=template's defaultDir, package
  manager=detected or npm, gitInit=false, modules=skipped
- Fails fast (instead of hanging) when the target directory already
  exists, instructing the caller to pass `--force`

All existing interactive behaviour is preserved when a TTY is present
and none of the new flags are set.


atinux:
Good to be reviewed @danielroe

packages/nuxi/src/commands/init.ts
```diff
@@ -104,6 +104,16 @@ export default defineCommand({
       type: 'string',
       description: 'Use Nuxt nightly release channel (3x or latest)',
     },
+    defaults: {
+      type: 'boolean',
+      alias: 'y',
+      description: 'Use defaults for all prompts (useful for CI/agent environments)',
+    },
+    interactive: {
```
posva: Currently `interactive` and `defaults` seem to only affect the same part of the code. I think `interactive` should not exist, it also makes it weird to have a default value of `true` (do you pass `--no-interactive`? `--interactive=false`?). I find the `-y` to be better

packages/nuxi/src/commands/init.ts
```diff
@@ -104,6 +104,16 @@ export default defineCommand({
       type: 'string',
       description: 'Use Nuxt nightly release channel (3x or latest)',
     },
+    defaults: {
```
posva: Should probably be named `yes`, it's more conventional (cf `npm init -h`)

packages/nuxi/src/commands/init.ts
```diff
@@ -139,26 +171,93 @@ export default defineCommand({
...
+          `  modules:        ${colors.cyan(effectiveModules)}`,
+        ].join('\n'),
+        'Available options',
+      )
+
+      // No project directory was given — the agent now has all the information
+      // it needs to re-run with explicit flags. Exit without creating anything.
+      if (!ctx.args.dir) {
+        outro('Re-run with a project directory to proceed, e.g: nuxi init <dir>')
+        process.exit(0)
```
posva: should be non-zero since the run didn't succeed, probably 2 because it's missing parts of the command but I don't think there is a standard on the actual value (other than being non-zero)

packages/nuxi/src/commands/init.ts
```diff
@@ -139,26 +171,93 @@ export default defineCommand({
...
+        ? '(none)'
+        : !ctx.args.modules
+            ? '(skipped)'
+            : ctx.args.modules as string
+
+      const templateLines = Object.entries(availableTemplates).map(([name, data]) =>
+        `  ${colors.cyan(name.padEnd(18))} ${data?.description ?? ''}${name === DEFAULT_TEMPLATE_NAME ? colors.dim('  ← default') : ''}`,
+      )
+
+      note(
```
posva, line 164: If you want to simplify this, you could just display the help, after that display  the template list (I think that's the only value that requires a specific value that isn't already mentioned in the help)

posva (commented):
This is what I think. Some agents automatically run `--help` to auto discover what they can do or they use it as soon as the command fails. So I really think you can keep this very lean in terms of changes as long as the `--help` text is good. I would say the only missing part are templates which can be listed if you detect an agent

packages/nuxi/src/commands/init.ts
```diff
@@ -116,6 +126,28 @@ export default defineCommand({
 
     intro(colors.bold(`Welcome to Nuxt!`.split('').map(m => `${themeColor}${m}`).join('')))
 
+    // Detect non-interactive environments: agent, CI, no TTY, or explicit flags
+    const isNonInteractive
+      = ctx.args.defaults === true
+        || ctx.args.interactive === false
+        || isAgent // AI coding agent (Claude, Cursor, Copilot, Devin, Gemini…)
```
danielroe: this will also be positive if a user is using a shell from within a coding harness.

I think it's better to gate this on whether the TTY is interactive or not, rather than whether an agent has been detected

packages/nuxi/src/commands/init.ts
```diff
@@ -139,26 +171,93 @@ export default defineCommand({
...
+          colors.bold('Proceeding with:'),
+          `  template:       ${colors.cyan(effectiveTemplate)}`,
+          `  directory:      ${colors.cyan(effectiveDir)}`,
+          `  packageManager: ${colors.cyan(effectivePM)}`,
+          `  gitInit:        ${colors.cyan(String(effectiveGitInit))}`,
+          `  install:        ${colors.cyan(String(effectiveInstall))}`,
+          `  modules:        ${colors.cyan(effectiveModules)}`,
+        ].join('\n'),
+        'Available options',
+      )
```
danielroe, line 173: this will quickly become out of date - can we instead harness citty's own tools?

additionally it's hard-coded to be a specific width (`padEnd(18)`), which is not really needed for agents and is also very dependent on the content of the

danielroe (commented):
I agree with @posva's comments. I've added a few of my own. I think we can make this much leaner/simpler. 🙏 

additionally we should add a unit test for this behaviour.

packages/nuxi/src/commands/init.ts
posva: I agree with this: `hasTTY` should be enough. Displaying the help for the setup to be run in one go is, IMO, the best option. It's worth noting that it will also render #1287 ineffective, but I believe that's the correct path for this kind of CLI because if by any chance we want later to support agent-interactive commands, e.g. `nuxi module add`, we can probably opt out `const isNonInteractive = !ctx.args.command && !hasTTY`

packages/nuxi/src/commands/init.ts
```diff
@@ -116,6 +121,14 @@ export default defineCommand({
 
     intro(colors.bold(`Welcome to Nuxt!`.split('').map(m => `${themeColor}${m}`).join('')))
 
+    const isNonInteractive = ctx.args.yes === true || !hasTTY
```
posva, line 124: Rather than defaulting immediately to `--yes` in non-TTY mode, if the option `--yes` isn't set, show the help so it can be invoked with the proper arguments.
Actually, it's a bit more complex, ideally the CLI should detect if it has all the information it needs to run without interactivity, if not, show the help and exit with non-zero (like 2)

packages/nuxi/src/commands/init.ts
```diff
@@ -104,6 +104,11 @@ export default defineCommand({
       type: 'string',
       description: 'Use Nuxt nightly release channel (3x or latest)',
     },
+    yes: {
+      type: 'boolean',
+      alias: 'y',
+      description: 'Use default values for all prompts',
```
posva:
```suggestion
      description: 'Use default values for and skip interactive prompts',
```

packages/nuxi/src/commands/init.ts
```diff
@@ -139,26 +152,55 @@ export default defineCommand({
...
+        `  ${colors.cyan(name)} ${data?.description ?? ''}${name === DEFAULT_TEMPLATE_NAME ? colors.dim(' (default)') : ''}`,
+      )
+
+      note(
+        [
+          'Available templates:',
+          '',
+          ...templateLines,
+          '',
+          `Run ${colors.cyan(`${initCmd} --help`)} for all options.`,
```
posva, line 170: I wonder if it should just display the help before the list of templates. I think this could be simplified by merging this message with my suggestion above: if options are missing show the help and the list of templates because it's needed to provide a value

danielroe:
marking draft as when we last spoke you were thinking of reworking or taking a totally different approach

unmark whenever you feel like it!

PR #1265, opened 2026-03-27
feat(cli): locks for agents running dev and build commands

harlan-zw:
### 🔗 Linked issue

https://github.com/nuxt/nuxt/pull/34629

### ❓ Type of change

- [ ] 📖 Documentation
- [ ] 🐞 Bug fix
- [ ] 👌 Enhancement
- [x] ✨ New feature
- [ ] 🧹 Chore
- [ ] ⚠️ Breaking change

### 📚 Description

Inspired by [Next.js 16.2's dev server lock file](https://nextjs.org/blog/next-16-2-ai#dev-server-lock-file) & https://github.com/nuxt/nuxt/pull/34629, this adds a `nuxt.lock` file to the build directory during `nuxi dev` and `nuxi build`. The lock file contains process info (PID, port, URL, command) so that a second invocation can detect the existing process and print an actionable error message with the running server's URL and a kill command.

This only runs for agents (using `std-env` agent detection). Agents can set `NUXT_IGNORE_LOCK=1` to bypass the check if they have a good reason to.

Stale locks from crashed processes are automatically cleaned up on the next startup via PID liveness checking (`process.kill(pid, 0)`). Pure Node.js, zero new dependencies.

#### Lock file format (`<buildDir>/nuxt.lock`)
```json
{
  "pid": 12345,
  "command": "dev",
  "port": 3000,
  "hostname": "127.0.0.1",
  "url": "http://127.0.0.1:3000",
  "startedAt": 1711540800000
}
```

#### Error output when a dev server is already running
```
Another Nuxt dev server is already running:

  URL:     http://127.0.0.1:3000
  PID:     12345
  Dir:     /path/to/project
  Started: 3/27/2026, 2:00:00 PM

Run `kill 12345` to stop it, or connect to http://127.0.0.1:3000
Set NUXT_IGNORE_LOCK=1 to bypass this check.
```

### Limitations

- These aren't OS-level file locks - we can't support these without native bindings. There are edge cases around force process exits and hards crashes but there is recover logic

### Future considerations

Rust bindings for native OS file locking.

### DIFF
--- a/packages/nuxi/src/commands/build.ts
+++ b/packages/nuxi/src/commands/build.ts
@@ -9,7 +9,7 @@ import { showVersions } from '../utils/banner'
 import { overrideEnv } from '../utils/env'
 import { clearBuildDir } from '../utils/fs'
 import { loadKit } from '../utils/kit'
-import { checkLock, formatLockError, writeLock } from '../utils/lockfile'
+import { acquireLock, formatLockError } from '../utils/lockfile'
 import { logger } from '../utils/logger'
 import { startCpuProfile, stopCpuProfile } from '../utils/profile'
 import { cwdArgs, dotEnvArgs, envNameArgs, extendsArgs, legacyRootDirArgs, logLevelArgs, profileArgs } from './_shared'
@@ -47,7 +47,7 @@ export default defineCommand({
       await startCpuProfile()
     }
 
-    let lockCleanup: (() => void) | undefined
+    let releaseLock: (() => void) | undefined
     try {
       intro(colors.cyan('Building Nuxt for production...'))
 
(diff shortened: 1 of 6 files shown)


packages/nuxi/src/commands/build.ts
```diff
@@ -94,6 +103,13 @@ export default defineCommand({
 
       await clearBuildDir(nuxt.options.buildDir)
 
+      // Write lock after clearing build dir so it doesn't get deleted
+      lockCleanup = await writeLock(nuxt.options.buildDir, {
+        pid: process.pid,
+        command: 'build',
+        startedAt: Date.now(),
+      })
```
danielroe: would it make sense to add `nuxt.lock` here so we don't clear it in the first place, rather than writing it again?

https://github.com/nuxt/cli/blob/ab4e0c6da3b69d45e800c473f2ee59a7ab7df146/packages/nuxi/src/utils/fs.ts#L22-L24

harlan-zw:
> I love the idea @harlan-zw
> 
> Do you think it would also be possible somehow, and would it be useful to also provide a way to read the logs for the current process?
> 
> EDIT: seems that Next.js has a path to a log file

Seems like a good idea, I think the scope of that may go beyond the locks to get it up and running properly. I'd say we can tackle that  in follow up PR.

danielroe (approved):

PR #1283, opened 2026-04-15
fix(dev): resolve package versions via their parents

BellYun:
### 🔗 Linked issue

Refs nuxt/nuxt#34809 — this covers the CLI side. The devtools side is in nuxt/devtools#986.

### 📚 Description

Resolves `vite` via `nuxt -> @nuxt/vite-builder -> vite` instead of resolving it directly from cwd.
Under pnpm 11's enableGlobalVirtualStore, vite is not reachable from CWD or nuxt itself. However, since @nuxt/vite-builder declares vite as a direct dependency, the lookup now works correctly through that path.
- Verified against a pnpm 11 globalVirtualStore repro.
- Added a regression test.

### DIFF
--- a/packages/nuxi/src/utils/banner.ts
+++ b/packages/nuxi/src/utils/banner.ts
@@ -17,8 +17,8 @@ export function getBuilder(cwd: string, builder: Exclude<NuxtOptions['builder'] 
     case '@nuxt/vite-builder':
     default: {
       const pkgJSON = getPkgJSON(cwd, 'vite')
-      const isRolldown = pkgJSON.name.includes('rolldown')
-      return { name: isRolldown ? 'Rolldown-Vite' : 'Vite', version: pkgJSON.version }
+      const isRolldown = pkgJSON?.name?.includes('rolldown') ?? false
+      return { name: isRolldown ? 'Rolldown-Vite' : 'Vite', version: pkgJSON?.version ?? '' }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


danielroe:
would you be able to go ahead and address the root cause? 🙏

I did some tinkering on the plane just now and I'm pretty sure we need to be able to resolve nuxt -> @nuxt/vite-builder -> vite instead of the current approach of directly trying to resolve vite.

BellYun:
> would you be able to go ahead and address the root cause? 🙏
> 
> I did some tinkering on the plane just now and I'm pretty sure we need to be able to resolve nuxt -> @nuxt/vite-builder -> vite instead of the current approach of directly trying to resolve vite.

It was just a stopgap. I felt it was urgent to stop the crash first since it's a pretty bad experience for users. Already looking into the nuxt -> @nuxt/vite-builder -> vite path now. 

I'll update this PR

BellYun:
@danielroe I've addressed the root cause by resolving via nuxt -> @nuxt/vite-builder -> vite instead of direct lookup.

Also added a regression test for pnpm 11 globalVirtualStore.

One more thing — it looks like the CodeQL check is stuck in "waiting for results" (possibly due to fork PR restrictions). Would you mind re-running it or taking a look?

Thanks!

danielroe (approved):
thank you ❤️

PR #1287, opened 2026-04-21
feat(cli): agent-mode support via once() and batch()

posva:
## Summary

Alias `@clack/prompts` to [`@posva/clack-prompts`](https://github.com/posva/clack/tree/feat/agent-mode) so that interactive commands keep working under AI-agent harnesses (Claude Code, Cursor, Aider) that re-run the script to answer each prompt through a JSON session file.

- Wrap non-idempotent side effects with `once(id, fn)` so they run exactly once across replays:
  - Template download + slug rename, nightly-version setup, install + `git init`, `nuxt module add` install + `nuxt.config` update.
  - The target-dir `existsSync` check — otherwise the second replay would see the folder the first run just created and diverge into the "directory already exists" branch.
- Collapse independent prompt pairs into `batch({...})` so agents answer them in one round-trip:
  - `init` — package-manager select + git-init confirm.
  - `upgrade` — nightly-channel select + dedupe/force/skip method select (only when both would fire).
- Stable `id`s on every prompt so clack can resolve answers deterministically across replays. Both batched and single-prompt fallbacks share the same ids so the same question maps to the same session entry regardless of which branch runs.
- `selectModulesAutocomplete` gains an optional `id` so the two call sites (`init`, `module add`) disambiguate, and no longer bails on `!hasTTY` when running under `isAgent()` — agents don't need a TTY.

Human UX is unchanged: `once` runs transparently in interactive mode, and `batch` displays its items sequentially just like `group`.

## Test plan

- [x] `pnpm test:types` — clean
- [x] `pnpm lint` — clean
- [x] `pnpm build` — all three buildable packages compile
- [ ] Interactive smoke: `nuxi init`, `nuxi upgrade`, `nuxi module add` each walk through prompts identically to today
- [ ] Agent smoke: run each command under an agent harness, verify that wrapped `once` steps execute once and batched prompts emit together

### DIFF
--- a/packages/nuxi/src/commands/init.ts
+++ b/packages/nuxi/src/commands/init.ts
@@ -5,7 +5,7 @@ import type { TemplateData } from '../utils/starter-templates'
 import { existsSync } from 'node:fs'
 import process from 'node:process'
 
-import { box, cancel, confirm, intro, isCancel, outro, select, spinner, tasks, text } from '@clack/prompts'
+import { batch, box, cancel, confirm, intro, isCancel, once, outro, select, spinner, tasks, text } from '@clack/prompts'
 import { defineCommand } from 'citty'
 import { colors } from 'consola/utils'
 import { downloadTemplate, startShell } from 'giget'
@@ -142,6 +142,7 @@ export default defineCommand({
     let templateName = ctx.args.template
     if (!templateName) {
       const result = await select({
+        id: 'init:template',
         message: 'Which template would you like to use?',
         options: Object.entries(availableTemplates).map(([name, data]) => {
           return {
@@ -173,6 +174,7 @@ export default defineCommand({
     if (dir === '') {
       const defaultDir = availableTemplates[templateName]?.defaultDir || 'nuxt-app'
       const result = await text({
+        id: 'init:dir',
         message: 'Where would you like to create your project?',
         placeholder: `./${defaultDir}`,
         defaultValue: defaultDir,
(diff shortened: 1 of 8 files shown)


danielroe:
did you see https://github.com/nuxt/cli/pull/1264 and do you have an opinion?

I imagine this is a test PR to evaluate how it works, before opening a PR/merging into clack/prompts?

packages/nuxi/src/commands/init.ts
```diff
@@ -498,7 +535,10 @@ export default defineCommand({
         ctx.args.logLevel ? `--logLevel=${ctx.args.logLevel}` : '',
       ].filter(Boolean)
 
-      await runCommand(addModuleCommand, args)
+      await once('init:add-modules', async () => {
+        await runCommand(addModuleCommand, args)
+        return null
+      })
```
posva, line 541: can be simplified to `() => runCommand()`

posva:
Yes, I saw #1264 . This is different: it's an experiment about bringing the interactivity of the CLI to agents so they can make choices based on instructions.

It's not meant to be merged

Oh, and I think #1264 is great and that one should def be shipped. Interactivety in CLIs should, at the very least fallback to showing the help when run by an agent, that would be so much better. Some times this is not possible: e.g. the dev server prompting the user to install a package (this is done in nuxt dev for example when using `@nuxt/content` it asks to install sqlite but it should just log it and not use an interactive prompt)

Issue #1300, opened 2026-04-30
Add `--portless` option for remote dev access


onmax:
It would be useful to support a `--portless` (similar to `--tunnel`), using [Portless](https://portless.sh) as an alternative way to expose a local Nuxt dev server over HTTPS. The goal would be a simple built-in option for sharing or testing a local app remotely without extra manual setup, similar to the existing tunnel workflow.

related https://github.com/unjs/listhen/issues/227

danielroe:
I think portless is meant to be used as the parent process, and it is pretty seamless!

```
portless myapp nuxt dev
```

I think adding a `--portless` flag wouldn't help us much.

PR #1301, opened 2026-04-30
feat(dev): add --portless support

onmax:
Adds `nuxi dev --portless`, precomputes the Portless hostname so Nuxt/Vite allowlists include it, and exposes the dev server through the external Portless CLI.


danielroe:
thank you for the PR ❤️ 

~> https://github.com/nuxt/cli/issues/1300#issuecomment-5077733130

PR #1307, opened 2026-05-07
fix: handle missing nuxt version in `readPackageJSON`

### LINKED ISSUE #1179: Error when initializing nuxt app with arguments
Hi,

I have a boilerplate that uses `npm create nuxt@latest` to initialize a Nuxt app, after which we add a set of components and other files. Our initialization script has recently stopped working.

Command:
```
npm create nuxt@latest -- test --template 'minimal' --gitInit --packageManager npm --modules 'eslint' --install false
```

Output:

```
    .d$b.
       i$$A$$L  .d$b
     .$$F` `$$L.$$A$$.
    j$$'    `4$$:` `$$.
   j$$'     .4$:    `$$.
  j$$`     .$$:      `4$L
 :$$:____.d$$:  _____.:$$:
 `4$$$$$$$$P` .i$$$$$$$$P`

┌  Welcome to Nuxt!
│
◇  Creating project in test
│
◇  Downloaded minimal template
│
●  Skipping install dependencies step.

 ERROR  Cannot find matching package.json in undefined or parent directories 
```

Is the command incorrect, or is this a bug?

Flo0806:
### 🔗 Linked issue

Fixes: #1179 

### 📚 Description

`npm create nuxt@latest -- test --template=minimal --modules=eslint --install=false`

With `--install=false`, the template is downloaded but `Nuxt` is never installed. `getNuxtVersion` then can't find it and throws an error. The error wasn't caught. Now we simple caught it and the process can finish correctly.

### DIFF
--- a/packages/nuxi/src/utils/versions.ts
+++ b/packages/nuxi/src/utils/versions.ts
@@ -6,7 +6,7 @@ import { coerce } from 'semver'
 import { tryResolveNuxt } from './kit'
 
 export async function getNuxtVersion(cwd: string, cache = true) {
-  const nuxtPkg = await readPackageJSON('nuxt', { url: cwd, try: true, cache })
+  const nuxtPkg = await readPackageJSON('nuxt', { url: cwd, try: true, cache }).catch(() => null)
   if (nuxtPkg) {
     return nuxtPkg.version!
   }


packages/nuxi/src/utils/versions.ts
```diff
@@ -6,7 +6,7 @@ import { coerce } from 'semver'
 import { tryResolveNuxt } from './kit'
 
 export async function getNuxtVersion(cwd: string, cache = true) {
-  const nuxtPkg = await readPackageJSON('nuxt', { url: cwd, try: true, cache })
+  const nuxtPkg = await readPackageJSON('nuxt', { url: cwd, try: true, cache }).catch(() => null)
```
danielroe, line 9: this seems like a bug, if so, because `try: true` should do this automatically

Flo0806: Ah, understand. I'll have a look.

danielroe (approved):
merging this as a hotfix until https://github.com/unjs/pkg-types/pull/263 can land.

thank you ❤️

PR #1328, opened 2026-05-28
fix(dev): timeout NuxtDevServer.close() to avoid dev-restart deadlock

mttzzz:
## Issue

Refs https://github.com/nuxt/nuxt/issues/32928

The Nuxt dev server can enter a permanent **"Restarting Nuxt..."** state after a `nuxt.config.ts` edit (or other change that triggers a full reload) when any nitro plugin holds a long-lived background handle.

`NuxtDevServer.close()` `await`s `currentNuxt.close()` which in turn awaits every registered `close` hook. If any plugin owns a connection that is blocked on a server response (Bull `BLPOP`/`BRPOPLPUSH`, Postgres `LISTEN`, native WebSocket, an `ioredis`-shared client, etc.) the close hook never resolves — and the restart deadlocks.

This is fairly easy to hit in real apps: queues (BullMQ/Bull), realtime endpoints, background pollers all rely on connections that don't drain in a few hundred ms.

## Change

Wrap the inner `close()` in a `Promise.race` against a 3-second timeout. After the deadline the restart proceeds; the new Nuxt instance binds the port (the OS reclaims the bind from the orphaned process) and the lingering handles are GC'd along with the dropped reference to the old Nuxt instance.

The timeout is overridable via `NUXT_DEV_CLOSE_TIMEOUT_MS` for projects that legitimately need more grace (e.g. heavy in-flight migrations on close).

```ts
async close(): Promise<void> {
  if (!this.#currentNuxt) return
  const timeoutMs = Number(process.env.NUXT_DEV_CLOSE_TIMEOUT_MS) || 3000
  let timer: NodeJS.Timeout | undefined
  await Promise.race([
    this.#currentNuxt.close().finally(() => { if (timer) clearTimeout(timer) }),
    new Promise<void>((resolve) => {
      timer = setTimeout(resolve, timeoutMs)
      timer.unref?.()
    }),
  ])
}
```

## Behaviour

- **Plugins that close cleanly** (the vast majority): no observable change — `close()` resolves in milliseconds, timer is cleared, no leftover handles in the event loop.
- **Plugins that block on a long-lived socket**: we wait `timeoutMs`, then proceed with the restart. Equivalent to the user pressing Ctrl-C after the deadlock and re-running `bun dev`, only automatic.

## Trade-offs

This is intentionally a safety net rather than a root-cause fix — the real fix lives in the user's plugin (`queue.close(true)`, `await server.unref()` on a websocket, etc.). But the failure mode is hard to diagnose for users (CLI says "Restarting Nuxt..." with no further output) and the timeout makes the dev experience consistent with what most people expect from an HMR-driven workflow.

## Tested

Reproduced locally with a Nuxt 4.4.6 + 6 Bull workers + Postgres LISTEN setup that previously deadlocked indefinitely on every `nuxt.config.ts` save. With the patch, restart completes in 6-10s consistently.

### DIFF
--- a/packages/nuxi/src/dev/utils.ts
+++ b/packages/nuxi/src/dev/utils.ts
@@ -508,7 +508,8 @@ export class NuxtDevServer extends EventEmitter<DevServerEventMap> {
     let timer: NodeJS.Timeout | undefined
     await Promise.race([
       this.#currentNuxt.close().finally(() => {
-        if (timer) clearTimeout(timer)
+        if (timer)
+          clearTimeout(timer)
       }),
       new Promise<void>((resolve) => {
         timer = setTimeout(resolve, timeoutMs)
--- a/packages/nuxi/src/dev/utils.ts
+++ b/packages/nuxi/src/dev/utils.ts
@@ -82,11 +82,16 @@ export const DEFAULT_CLOSE_TIMEOUT_MS = 3000
  * (success or rejection — we don't want a rejected nuxt close to abort the
  * subsequent restart) or when the timer fires, whichever happens first.
  * Exposed for testing; intended to be called from `NuxtDevServer.close()` only.
+ *
+ * The closer is wrapped in `Promise.resolve().then(closer)` so a synchronous
+ * throw from the closer is also rerouted through `.catch` and cannot abort
+ * the restart.
  */
 export async function closeWithTimeout(closer: () => Promise<void>, timeoutMs: number): Promise<void> {
   let timer: NodeJS.Timeout | undefined
   await Promise.race([
-    closer()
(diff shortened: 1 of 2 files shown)


mttzzz:
Added a unit test for the extracted `closeWithTimeout` helper (5 specs covering the fast-close, never-resolving closer, rejection, and timer-cleanup paths). Should bring patch coverage to 100% and re-trigger the CodeRabbit review that was aborted by the autofix commit.

danielroe:
I think this has been resolved in https://github.com/nuxt/cli/pull/1424 - you can try `@nuxt/cli-nightly` - does it solve your issue? 🙏

PR #1330, opened 2026-06-03
fix(init): use template's package manager instead of prompting

### LINKED ISSUE #1187: nuxt ui template already set the package manager
The nuxt ui template has been set to pnpm before choose the package manager.

https://github.com/nuxt-ui-templates/starter/blob/638e860fc2f249acbe1f7a73fa43283ecb0605cd/package.json#L25

Image

benjamincanac:
## Summary

Closes #1187.

Some starter templates (e.g. the Nuxt UI starter) ship pinned to a package manager — a `packageManager` field in `package.json` (like `pnpm@x`), a lockfile (`pnpm-lock.yaml`) and/or workspace/config markers (`pnpm-workspace.yaml`, `.yarnrc.yml`). The `nuxi init` package manager prompt is seeded from `npm_config_user_agent`, not the template, so picking a different one leaves the project in a broken state: a stale lockfile, a mismatched `packageManager` field, and workspace config that no longer applies.

Rather than cloning the template and stripping those files after the fact (which can silently break a template that relies on its workspace config), this respects the template's choice:

- detects the template's package manager via nypm's `detectPackageManager`, scoped to the template dir (`includeParentDirs: false`, `ignoreArgv: true`)
- when the template pins one, uses it and **skips the package manager prompt** — the template is left untouched
- an explicit `--packageManager` still takes precedence, with a warning when it conflicts with the template's pin (its lockfile/config may no longer match)
- when the template pins nothing, behaves as before (prompt, or use `--packageManager`)

This makes shipping a template that works across package managers a choice for the template author (ship it without a lockfile), as discussed in the PR thread.

## Test plan

- [x] reworked `packages/nuxi/test/unit/commands/init.spec.ts` to cover detection from a lockfile, from the `packageManager` field, and the no-pin case
- [x] `pnpm test`


atinux (approved):
Approving as I was involved in pair-programming on this one 😊

danielroe:
is there a reason you don't want to remove the lockfile from the template?

if it is because it contains workspace config

benjamincanac:
@danielroe not sure to understand, it does remove the lockfile 🤔 https://github.com/nuxt/cli/pull/1330/changes#diff-02bf0f3c4f7503f4ebacc612fc62c0c6fc3d2e6fb806b5a52afdc72cd939a062R624

danielroe:
I mean, in the repository which is the template, rather than in nuxt/cli.

benjamincanac:
Well yes all Nuxt UI templates are deployed on Vercel: https://github.com/nuxt-ui-templates/dashboard. I use renovate to keep the deps up-to-date on a weekly basis to make sure they continue working. It wouldn't be possible without a lockfile.

danielroe:
it feels like the right fix is to ship a template without a lockfile in the first place, rather than remove it after the fact when nuxt/cli clones it.

I understand you deploy it, and that's great, but how about deploying it without a lockfile? (that way you're always on the latest versions?) or you could have a separate branch or folder which didn't contain the lockfile.

but it's about more than the lockfile. for example, if your `pnpm-workspace.yaml` contains workspaces configuration, the template would be silently broken after install when this is gone. properly speaking we would need to adapt this configuration to npm workspaces or bun workspaces.

the same might be true for other things and makes me wonder if we're approaching this the wrong way round by 'fixing' it in nuxt/cli. even if we make it work for that particular nuxt/ui template, I'm not sure it would be desired behaviour in general

benjamincanac:
I don't think it's a good idea to ship a template without a lockfile, first of all it's not practical in development and since any release can break it is not a good experience for the user. It is fine for a starter with a `nuxt` dependency but not for real world applications like https://chat-template.nuxt.dev/ for example.

This PR only removes the lockfile when the user chooses a different package manager than the one defined in the template.

Should I close this then?

atinux:
Actually, most of the templates listed on https://nuxt.com/templates or https://hub.nuxt.com/templates also have lock files as this is the most convenient for any developers building a template.

The case of a very specific workspace is very rare IMO, but we could display a `(recommended)` if we detect the lock file in the template but let the user chose a different package manager, would that work @danielroe ?

danielroe:
I think the right solution is to ship templates without lockfiles, rather than normalising it after the fact at clone time ...

if you want to deploy or test them with lockfiles, I recommend having templates located in subfolders of a repo, with the lockfile _outside_ the template folders.

alternatively, we could simply avoid asking the user what package manager they want if one is already configured in the template they are cloning.

atinux:
> I think the right solution is to ship templates without lockfiles, rather than normalising it after the fact at clone time ...

There may be a gap in understanding around our ecosystem, how demos are built, and how quickly we want the community to be able to use the template/demo.

Some of our demos/templates have lockfiles: [nuxt/hackernews](https://github.com/nuxt/hackernews), [nuxt/movies](https://github.com/nuxt/movies)

If the issue is about deleting the files, in that case I think the potential solutions are:
1. avoid asking the user the package manager, let hir deal with it after the installation
2. have the possibility to clone repos and filter out the lock files (idea from @pi0)

danielroe:
I think the right approach for now is to avoid prompting for package manager choice if lock files are present in the template.

then it is a choice on the provider of the template whether they want to provide a template that works in different package managers - or not.

benjamincanac:
@danielroe I've applied your suggestions!

danielroe (approved):
❤️

PR #1389, opened 2026-07-25
fix(_dev,start): hide `_dev` and `start` commands from help output

danielroe:
### 📚 Description

this is confusing for users


packages/nuxt-cli/src/commands/dev-child.ts
```diff
@@ -9,6 +9,9 @@ export default defineCommand({
   meta: {
     name: '_dev',
     description: 'Run Nuxt development server (internal command to start child process)',
+    // `@nuxt/test-utils` spawns `nuxi _dev`, so the command stays runnable but is
+    // kept out of help output and completions
```
danielroe:
```suggestion
```

PR #1395, opened 2026-07-26
fix(dev): `stdin` paused when using clark spinners on Windows

userquin:
### 📚 Description

When using clark spinners stdin requires 2 enters, for example, pressing `h + enter` on Windows after installing cloudflare executable requires an additional `h + enter`, the `stdin` being paused (`nuxi dev --tunnel`).


danielroe (approved):
🔥

PR #1397, opened 2026-07-26
feat: support pnpm catalogs

### LINKED ISSUE #866: add support for `pnpm` catalogs
[`pnpm` catalogs](https://pnpm.io/catalogs) are a great way to manage dependencies that need to be kept in sync. Since the config is stored in `pnpm-workspace.yaml`, users can benefit from yaml's anchors and aliases to define these values only once (more details here: https://smcleod.net/2022/11/yaml-anchors-and-aliases/).

I recently opened a PR for the `nuxtr` vscode extension to allow it to resolve `nuxt`'s version when using catalogs: https://github.com/nuxtrdev/nuxtr-vscode/pull/185

A similar approach could be applied here but I believe it would be better to make the changes upstream in `pkg-types` and/or `nypm` so the whole ecosystem can benefit from the update. Plus that approach is incredibly naive and lacks tests to ensure proper resolution for all possible types of configs.

I know @antfu has done some work on his projects to start supporting catalogs, so he probably has a better understanding of the things we're missing to get this done.

danielroe:
### 🔗 Linked issue

resolves https://github.com/nuxt/cli/issues/866

### 📚 Description

this follows path taken by https://github.com/antfu-collective/ni in resolving/handling pnpm catalogs

### DIFF
--- a/packages/nuxt-cli/package.json
+++ b/packages/nuxt-cli/package.json
@@ -67,6 +67,7 @@
     "perfect-debounce": "^2.1.0",
     "picocolors": "^1.1.1",
     "pkg-types": "^2.3.1",
+    "pnpm-workspace-yaml": "^1.6.1",
     "rc9": "^3.0.1",
     "scule": "^1.3.0",
     "source-map": "^0.7.6",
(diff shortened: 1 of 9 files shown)


danielroe:
@coderabbitai review

Issue #994, opened 2025-07-29
ERROR  [unhandledRejection] read ECONNRESET in dev environment


sebaguse:
### Environment

- Operating System: `Windows_NT`
- Node Version:     `v22.14.0`
- Nuxt Version:     `4.0.2`
- CLI Version:      `3.27.0`
- Nitro Version:    `2.12.4`
- Package Manager:  `npm@11.4.2`
- Builder:          `-`
- User Config:      `app`, `runtimeConfig`, `$development`, `css`, `modules`, `quasar`, `pinia`, `imports`, `vite`, `nitro`, `devServer`, `ssr`, `experimental`, `build`, `alias`, `piniaPluginPersistedstate`, `fonts`, `security`, `compatibilityDate`, `devtools`, `future`, `pwa`
- Runtime Modules:  `nuxt-quasar-ui@2.1.13`, `@pinia/nuxt@0.11.2`, `@vueuse/nuxt@13.6.0`, `pinia-plugin-persistedstate/nuxt@4.4.1`, `@formkit/auto-animate/nuxt@0.8.2`, `@nuxt/eslint@1.7.1`, `@nuxt/fonts@0.11.4`, `@vite-pwa/nuxt@1.0.4`
- Build Modules:    `-`

### Reproduction

Every time I start a dev environment using `nuxi dev`, I encounter an error that causes the dev server to stop responding and the build to restart. I have to refresh the application with `Ctrl+F5` to get it working. It seems to run fine after that. 

This issue occurs since I updated my application to Nuxt v4. 

Here's the error message:
``` 
ERROR  [unhandledRejection] read ECONNRESET                                                                                                                                                                    11:02:00  

    at TLSWrap.onStreamRead (node:internal/stream_base_commons:216:20)
    at TLSWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

### Describe the bug

See: Reproduction.

### Logs

```shell-script

```

https://github.com/nuxt-modules/robots/issues/221

danielroe:
we need to investigate.

for now, can you run with `NUXT_SOCKET=0` to see if that resolves it?

I'm currently without a windows environment but there seems to be a regression with sockets on windows.

Vincentdevreede:
https://github.com/nuxt/nuxt/issues/23832#issuecomment-3143391239

sebaguse:
[...]

Issue #247, opened 2023-10-11
ERROR  [unhandledRejection] read/write ECONNRESET


cdwmhcc:
### Environment

------------------------------
- Operating System: Windows_NT
- Node Version:     v20.8.0
- Nuxt Version:     3.7.4
- CLI Version:      3.9.0
- Nitro Version:    2.6.3
- Package Manager:  pnpm@8.9.0
- Builder:          -
- User Config:      devtools
- Runtime Modules:  -
- Build Modules:    -
------------------------------

### Reproduction

1. `npx nuxi@latest init my-app`
2. select pnpm
3. init git
4. cd my-app
5. pnpm dev

Open http://localhost:3000, refresh the web page quickly or press `ctrl + R` continuously.

### Describe the bug

``` node
 ERROR  [unhandledRejection] write ECONNABORTED                                                               13:33:13

  at afterWriteDispatched (node:internal/stream_base_commons:160:15)
  at writeGeneric (node:internal/stream_base_commons:151:3)
  at Socket._writeGeneric (node:net:952:11)
  at Socket._write (node:net:964:8)
  at writeOrBuffer (node:internal/streams/writable:447:12)
  at _write (node:internal/streams/writable:389:10)
  at Writable.write (node:internal/streams/writable:393:10)
  at Socket.ondata (node:internal/streams/readable:817:22)
  at Socket.emit (node:events:514:28)
  at addChunk (node:internal/streams/readable:376:12)


 ERROR  [unhandledRejection] read ECONNRESET                                                                  13:33:16

  at TCP.onStreamRead (node:internal/stream_base_commons:217:20)


 ERROR  [unhandledRejection] read ECONNRESET                                                                  13:33:16

  at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
```

![image](https://github.com/nuxt/nuxt/assets/8030134/9c2b2260-0eb4-4986-a37d-dcf28bd6259c)

### Additional context

Tried nodejs 18.17.1 18.18.0 18.18.1

daniluk4000:
@cdwmhcc did you manage to solve this?

S1ipKn0T:
Any response for this issue? I have this issue too with my localhost:300 and idk why when I switch between some pages this error occurred!

shehab-work:
Same issue here for Nuxt 4.0

Ahmed-seyam:
same issue here

PR #1412, opened 2026-07-30
fix(dev): do not restart when a client aborts a connection

kedrzu:
### Problem

`nuxt dev` can restart itself in a loop with:

```
 ERROR  [unhandledRejection] read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:216:20)
    at TCP.callbackTrampoline (node:internal/async_hooks:130:17)

●  Restarting Nuxt due to error: Error: read ECONNRESET
```

A client hanging up mid-request is ordinary dev-server traffic — a browser reloading while a page is still streaming, a tab closed mid-navigation, a stale client from another project reconnecting to a port this server now owns. When such an error surfaces as an unhandled rejection, both places that watch for one treat it as fatal:

- `createRestartHook` schedules a hard restart;
- the forked dev process reports it over IPC and calls `process.exit()`.

With a client that keeps reconnecting, this loops until the dev server is unusable — every reconnect kills the session that was just rebuilt.

### Where the rejection comes from

I traced the origin in a real project by wrapping the candidate call sites and logging which one rejects. It is `NodeDevWorker.handleUpgrade` in nitro, which returns `proxy.ws(...)` without awaiting or catching it:

```js
// nitropack/dist/core/index.mjs
handleUpgrade(req, socket, head) {
  if (!this.ready) { return }
  return this.#proxy.proxy.ws(req, socket, { target: this.#address, xfwd: true }, head)
}
```

When the Nitro dev worker is replaced, every proxied websocket dies and that promise rejects. Adding a `.catch()` there removed the rejections entirely (8 in 30s → 0) and the restarts with them, so the leak itself belongs in nitro and I am happy to open that separately.

This PR is the guard on the other side: whatever leaks it, a connection the peer dropped should not end the dev session.

### Change

`isBrokenPipe` already encodes exactly this rule — *"a property of the other end of the pipe, not a fault in the dev server"*. This adds a sibling predicate for aborted connections (`ECONNRESET`, `ECONNABORTED`, `ERR_STREAM_PREMATURE_CLOSE`) and applies it in both fatal paths.

The IPC handler also moves from `once` to `on`, so an ignored connection error no longer consumes the listener and leaves a later genuine rejection unreported.

Logging is deliberately untouched: the error still reaches the reporter, so the event stays visible in the terminal — it just no longer ends the session.

### Tests

- `errors.spec.ts` — coverage for the new predicate.
- `restart-hook.spec.ts` — an `ECONNRESET` no longer triggers the restart callback, while an unrelated error still does.

`pnpm lint`, `pnpm test:types` and `pnpm test:knip` are clean. `pnpm test:unit` passes except for two pre-existing failures in `test/e2e/commands.spec.ts` (`dev`, `works with dev server`, both expecting 200 and getting 404) — I confirmed those fail identically on an unmodified `main` in my environment, so they are unrelated to this change.

### DIFF
--- a/packages/nuxt-cli/src/dev/index.ts
+++ b/packages/nuxt-cli/src/dev/index.ts
@@ -6,7 +6,7 @@ import type { NuxtDevContext, NuxtDevIPCMessage, NuxtParentIPCMessage } from './
 import process from 'node:process'
 import defu from 'defu'
 import { overrideEnv } from '../utils/env.ts'
-import { isAbortedConnection, isBrokenPipe } from '../utils/errors'
+import { isRemotePeerError } from '../utils/errors'
 import { debug } from '../utils/logger'
 import { startCpuProfile, stopCpuProfile } from '../utils/profile.ts'
 import { openInspector } from './inspect'
@@ -25,8 +25,8 @@ function formatErrorMessage(error: unknown): string {
  */
 export function createRejectionHandler(report: (message: string) => void, stop: () => void): (reason: unknown) => void {
   return (reason: unknown) => {
-    if (isAbortedConnection(reason)) {
-      debug('Ignoring aborted connection:', reason)
+    if (isRemotePeerError(reason)) {
+      debug('Ignoring remote peer error:', reason)
       return
     }
     report(formatErrorMessage(reason))
@@ -53,8 +53,6 @@ class IPC {
       process.once('disconnect', () => {
         process.exit(0)
       })
-      // `on` rather than `once`, so that an ignored connection error does not
-      // consume the listener and leave a later genuine rejection unreported.
       process.on('unhandledRejection', createRejectionHandler(
         message => this.send({ type: 'nuxt:internal:dev:rejection', message }),
         () => process.exit(),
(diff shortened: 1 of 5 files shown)


kedrzu:
Two follow-ups since opening this.

**Codecov.** The flagged lines were the branch inside the `IPC` constructor, which no test could reach — `IPC` is instantiated at import time and only arms its listener inside a fork. Pushed `fa6fa33`, which lifts the decision into `createRejectionHandler` next to the existing `createRestartHook` and takes the report/stop side effects as arguments. The constructor keeps only the wiring, and both branches are now unit-tested.

**Where the rejection comes from.** I traced the origin rather than guessing: it is `NodeDevWorker.handleUpgrade` in Nitro, which returns `proxy.ws(...)` without awaiting or catching it, so replacing the dev worker rejects it with `ECONNRESET`. Filed as nitrojs/nitro#4493 with a fix in nitrojs/nitro#4494.

That does not make this PR redundant. It closes one leak; this closes the class. Whatever leaks a dropped connection, it should not end a dev session — and this repo already says so in `isBrokenPipe`: *"a property of the other end of the pipe, not a fault in the dev server"*. This just extends that to connections the peer aborted.

It may also explain some history here: #247, #968 and #994 all report this same `ECONNRESET` restart and were closed as completed, but the code paths that make it fatal are unchanged on `main` today. #247's reproduction — *"refresh the web page quickly or press ctrl+R continuously"* — is exactly a client dropping requests mid-flight.

danielroe (approved):
thank you ❤️

PR #1414, opened 2026-07-31
feat(dev): take over a dev server already running on this project 

danielroe:
### 🔗 Linked issue

closes https://github.com/nuxt/cli/pull/1337 (and incorporates some of it ❤️)

### 📚 Description

this PR does a number of things. it enables the lock file (https://github.com/nuxt/cli/pull/1265) whether or not we're using an agent, and additionally records whether or not the dev server was started in an interactive console.

<details><summary><b>Video preview</b></summary>
<p>

https://github.com/user-attachments/assets/2d936e0f-8645-41f7-a739-7cf3a26613c0

</p>
</details>

we then are able to protect corruption of `.nuxt/` and `.output/` with the following rules:

| running | starting | what happens |
| --- | --- | --- |
| dev | build | both run happily |
| build | dev | refused if build was started when no `.nuxt/` dir existed |
| build | build | refused |
| dev | dev | see below |

if you start a second dev process, it's now possible for it to _take over_ the existing port, in certain conditions (based on interactivity):

| running | starting | what happens |
| --- | --- | --- |
| non-interactive | non-interactive | takes over, logged loudly |
| non-interactive | interactive | prompts, default `takeover` |
| interactive | non-interactive | refuses, exits `1`, prints where the other server is and that `--takeover` forces it |
| interactive | interactive | prompts, default `don't start` |

this is designed to allow agents to take over easily, but _not_ kill a dev server that a person is operating. and also to handle the 'too many terminals' syndrome that I have 😭 

takeover is additionally restricted by the following conditions:
- a lockfile has to exist
- the previous process pid still has to exist (if not, we ignore the lockfile)
- the previous port has to still be in use (if not, we ignore the lockfile)
- the 'target' port for both process has to be the same (if not, we refuse to start)

before taking over, we write the pid of the new process into the lockfile (so the outgoing dev server can blame why it's going silently into the dark night) and then kill it, first politely and then non-negotiably. but if that fails, we give an informative error and die.

### DIFF
--- a/packages/nuxt-cli/src/dev/takeover.ts
+++ b/packages/nuxt-cli/src/dev/takeover.ts
@@ -72,8 +72,8 @@ export async function takeOverDevServer(buildDir: string, options: TakeoverOptio
   }
 
   // Signalling a `build` is never on the table, and neither is a server on a
-  // port we were not asked for: an explicit `--port` that differs means the user
-  // wants a second server.
+  // port we were not asked for: an explicit `--port` that differs skips the
+  // takeover, and the ordinary lock check decides whether starting is allowed.
   if (existing.command !== 'dev' || !existing.port) {
     return { action: 'none' }
   }
--- a/packages/nuxt-cli/src/dev/takeover.ts
+++ b/packages/nuxt-cli/src/dev/takeover.ts
@@ -1,0 +1,233 @@
+import type { LockInfo } from '../utils/lockfile'
+
+import process from 'node:process'
+
+import { cancel, isCancel, select } from '@clack/prompts'
+import { checkPort } from 'get-port-please'
+import colors from 'picocolors'
+
+import { restoreRawMode } from '../utils/console'
+import { isInteractiveSession, isLockEnabled, markTakenOver, readLock } from '../utils/lockfile'
+import { logger } from '../utils/logger'
+
+/** How long the outgoing dev server has to exit and release its port. */
+const TAKEOVER_TIMEOUT_MS = 5000
+/** How long a `SIGKILL`ed process has to disappear before we give up. */
+const TAKEOVER_KILL_TIMEOUT_MS = 1000
+const TAKEOVER_POLL_INTERVAL_MS = 100
+
+export type TakeoverRefusalReason
+  /** The holder was started in a terminal and we cannot ask this user. */
+  = | 'holder-interactive'
+  /** The user declined at the prompt. */
+    | 'declined'
+  /** `--no-takeover` was passed. */
+    | 'disabled'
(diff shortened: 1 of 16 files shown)


packages/nuxt-cli/src/dev/takeover.ts
```diff
@@ -0,0 +1,280 @@
...
+  // On Windows `SIGTERM` is not delivered as a signal and terminates the process
+  // outright, so the graceful window below simply passes quickly there.
+  signalAll(pids, 'SIGTERM')
+  if (await waitForRelease(pids, port, existing.hostname, timeouts.graceful ?? TAKEOVER_TIMEOUT_MS)) {
+    progress.stop(`Stopped the dev server on port ${port} (PID ${existing.pid})`)
+    return { action: 'taken', port, pid: existing.pid }
+  }
+
+  progress.update(`Waiting for the dev server on port ${port} to exit`)
+  signalAll(pids, 'SIGKILL')
```
TheAlexLichter, line 163: Do we need a solution like https://github.com/nuxt/test-utils/pull/1760 here?

packages/nuxt-cli/src/utils/lockfile.ts
```diff
@@ -61,17 +168,19 @@ function isLockActive(info: LockInfo): boolean {
 }
 
 /**
- * Locking is enabled for agents by default. `NUXT_LOCK=1` forces it on for
- * non-agents; `NUXT_IGNORE_LOCK=1` forces it off.
+ * Locking is always enabled: `dev` and `build` share one build directory, and
+ * two processes writing it corrupt each other's output. `NUXT_IGNORE_LOCK=1`
+ * and `NUXT_LOCK=0` opt out.
  */
 export function isLockEnabled(): boolean {
```
TheAlexLichter, line 175: should we always return `false` in CI?

packages/nuxt-cli/src/utils/lockfile.ts
```diff
@@ -1,25 +1,47 @@
...
+// build clears, and already a Nuxt-owned cache location.
+const OUTPUT_LOCK_DIRNAME = 'node_modules/.cache/nuxt'
 // PID recycling safety net. Locks older than this cannot be trusted because a
 // recycled PID could match a dead build's record.
 const MAX_LOCK_AGE_MS = 24 * 60 * 60 * 1000
 
-function isProcessAlive(pid: number): boolean {
+/** Whether this process is attached to a terminal a user can answer prompts on. */
+export function isInteractiveSession(): boolean {
+  return !!process.stdin.isTTY && !isCI
```
TheAlexLichter, line 41: worth checking for `isAgent`?

PR #1417, opened 2026-07-31
feat(curl,task): add `nuxt curl`, `nuxt task list` and `nuxt task run`

### LINKED ISSUE #353: Integrate with Nitro Tasks
Nitro introduced new (experimental) [tasks API](https://nitro.unjs.io/guide/tasks#using-cli).

In the future it should be possible to directly run tasks using CLI (`nuxi task run db:migrate`)

This feature is in early stages and yet, requires dependency on a **running dev server**. 

I plan to introduce it for Nuxi CLI but perhaps only after we don't have this limitation anymore.

danielroe:
### 🔗 Linked issue

closes #353 

### 📚 Description

this picks up on the lockfile support added in https://github.com/nuxt/cli/pull/1414 to add two more commands that use the lock file to find the running server:

<details><summary><b>Video preview</b></summary>
<p>

https://github.com/user-attachments/assets/5072f271-d75a-42f3-a57f-b65bfbb9488a

</p>
</details>

#### `nuxt curl`

> sends requests to your dev server without needing to know the port

```bash
nuxt curl /api/hello
nuxt curl /api/users -X POST -d '{"name":"daniel"}'
nuxt curl /api/data -H "Authorization: Bearer token"
```

it auto-detects JSON bodies, pretty-prints JSON responses in a TTY, and exits `22` on HTTP errors (matching real curl's `--fail`).

#### `nuxt task list` and `nuxt task run`

> these commands talk to Nitro's task runner 

```bash
# shows descriptions and any scheduled crons
nuxt task list

# prints the task's result and exits non-zero with the server's error when the task fails
nuxt task run db:migrate --payload.force=true
nuxt task run db:migrate --payload '{"force":true}'
```

> [!IMPORTANT]
> `--payload.key=value` pairs are always strings, as they are when Nitro reads them from the query string, so the JSON form is an escape hatch.

> [!NOTE]
> `nuxt task` works with nitropack 2 (nuxt 4) _and_ Nitro 3 (nuxt 5)

behind the scenes, both commands resolve the server in this order:

| source | when |
| --- | --- |
| `--url` flag | always wins |
| lock file (`nuxt.lock`) | if a dev server is running |
| Nitro dev worker socket | `nuxt task` only, when there is no lock† |

† the worker fallback means `nuxt task` works even when the dev server was started by something other than `nuxt dev` (for example, programmatically from nuxt/kit).

### DIFF
--- a/packages/nuxt-cli/src/commands/curl.ts
+++ b/packages/nuxt-cli/src/commands/curl.ts
@@ -1,0 +1,203 @@
+import { Buffer } from 'node:buffer'
+import { readFile } from 'node:fs/promises'
+import process from 'node:process'
+import { styleText } from 'node:util'
+
+import { defineCommand } from 'citty'
+
+import { findDevServer, noDevServerMessage } from '../utils/dev-server'
+import { logger } from '../utils/logger'
+import { logNetworkError } from '../utils/network'
+import { resolveRootDir } from '../utils/paths'
+import { rootDirArgs } from './_shared'
+
+const HAS_SCHEME_RE = /^[a-z][a-z\d+.-]*:\/\//i
+const JSON_TOKEN_RE = /("(?:\\.|[^"\\])*")(\s*:)?|\b(?:true|false|null)\b|-?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?/g
+const JSON_CONTENT_TYPE_RE = /^application\/(?:[\w.+-]+\+)?json\b/i
+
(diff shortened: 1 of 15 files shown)


TheAlexLichter:
Somewhat (un)related note as I saw `nuxt.lock`: Vite might get a lockfile themselves (https://github.com/vitejs/vite/issues/23065#issuecomment-5089296404)

danielroe:
nice! it makes sense.

I am assuming vite wouldn't write a lockfile if it's being used programmatically by something like nuxt

PR #1432, opened 2026-08-05
feat(curl,task): highlight + pretty-print responses

danielroe:
### 📚 Description

this follows up on https://github.com/nuxt/cli/pull/1417 to highlight more responses (currently JSON + HTML). we use https://github.com/speed-highlight/core for this. it's a well-maintained library that's tiny (when inlined - 9.3 kB). only issue is that it doesn't expose `tokenize` - opening a PR shortly to add this feature...

we also pretty-print html, only rewriting whitespace at block-level boundaries to avoid changing the meaning of a request.

| before | after |
| - | - |
| screenshot of output of `nuxt curl /` before this PR | screenshot of output of `nuxt curl /` after this PR |


atinux (approved):

atinux:
Could be nice to highlight markdown as it becomes a new standard for LLM to send back Markdown with specific Accept header

PR #1532, opened 2026-09-10
fix(dev): quit command hangs/timeouts on Windows

### LINKED ISSUE #1531: alpha version: quit command hangs/timeouts on Windows 10 PRO
Can be tested with the playground here (node v24.19.0):
- `nr dev` at root (one terminal)
- `pnpm nuxt dev` (in another terminal)
- open / and navigate to /ws, then click ping button
- on playground terminal press q: the process awaits the default timeout (15s) and then logs: 
```shell
[nitro-runtime] close: 0.054ms
◐  Cleaning up... press Ctrl-C again to exit immediately..
 WARN  force closing dev worker...                                                                                                                                                

◒  Cleaning up... press Ctrl-C again to exit immediately...function                                                                                                                 
│
▲  The dev server did not shut down within 15s: 80 file watchers, 1 worker thread, 1 open connection, 2 timers. Exiting anyway.
◇  Canceled                                                                           
```

I added a hack to show the opened connections adding the following code before calling `summariseActiveResources` at `setupSignalHandlers` in [dev.ts](https://github.com/nuxt/cli/blob/main/packages/nuxt-cli/src/commands/dev.ts#L469):
```ts
        const handles = (process as any)._getActiveHandles?.()
        const networkHandles = handles.filter(h => (
          h.constructor.name === 'Socket'
          || h.constructor.name === 'Server'
        ))
        console.dir(networkHandles, { depth: 2 })
        const summary = summariseActiveResources()
```

Moving [this.#websocketConnections.add(socket)](https://github.com/nuxt/cli/tree/main/packages/nuxt-cli/src/dev/utils.ts#L1137) before the if statement to allow [#closeWebSocketConnections](https://github.com/nuxt/cli/tree/main/packages/nuxt-cli/src/dev/utils.ts#L1289-L1294) to close also these connections, the `1 open connection` dissapear.

I'm working on fixing file watchers...

userquin:
### 🔗 Linked issue

resolves #1531 

### 📚 Description

This PR includes:
- track Vite HMR connections to allow close them, check https://github.com/nuxt/cli/pull/1532#issuecomment-5626405787
- WIP: resolve deadlock: upstream, check https://github.com/nuxt/cli/pull/1532#issuecomment-5626202941

Maybe we can fix the deadlock here, not yet sure...

### DIFF
--- a/packages/nuxt-cli/src/dev/utils.ts
+++ b/packages/nuxt-cli/src/dev/utils.ts
@@ -1129,15 +1129,16 @@ export class NuxtDevServer extends EventEmitter<DevServerEventMap> {
         const baseURL = nuxt.options.app.baseURL.startsWith('./') ? nuxt.options.app.baseURL.slice(1) : nuxt.options.app.baseURL
         const assetsDir = nuxt.options.app.buildAssetsDir
         const viteHmrPath = `${baseURL.replace(/\/$/, '')}/${assetsDir.replace(/^\//, '')}`
-        if (req.url?.startsWith(viteHmrPath)) {
-          return // Skip for Vite HMR
-        }
-        nuxt.server.upgrade(req, socket as any, head)
-
+        // fix #1531: track all opened connections
+        // quit command hangs/timeouts on Windows 10 with Vite HMR
         this.#websocketConnections.add(socket)
         socket.on('close', () => {
           this.#websocketConnections.delete(socket)
         })
+        if (req.url?.startsWith(viteHmrPath)) {
+          return // Skip for Vite HMR
+        }
+        nuxt.server.upgrade(req, socket as any, head)
       })
     }
 


userquin:
Nuxt cli timeout: there is a bug at nitropack (nitro v2.13.4), when closing the worker, if there is a timeout, the caller will hang forever, in this case, nuxt cli will timeout with its default (15 seconds), missing `resolve` call on `setTimeout` at [#closeWorker](https://github.com/nitrojs/nitro/blob/v2/src/core/dev-server/worker.ts#L150-L154). 

The `#closeWorker` is fine at nitro v3.

Patching nitropack on my local node_modules calling the `resolve` method inside the `setTimeout`:
teemirnal showing nuxt cli closes gracefully with the nitropack patch but with 5 seconds of time

userquin:
Found the issue, it is a dead lock with nitropack 🥳 🥳 🥳

userquin:
### The dead lock

When shutting down the dev server (nitro dev-server at nitropack), the process often delays for 5 seconds until it throws `WARN force closing dev worker...`, ultimately failing to gracefully close the IPC sockets.

### Root Cause

There is a deadlock between the dev worker and the parent proxy during shutdown. 
Inside the dev worker's shutdown sequence, it executes `await new Promise((resolve) => listener?.close(resolve))`. The `listener.close()` method waits for all active connections to drain before resolving. However, the parent process holds keep-alive/upgrade connections open and waits for the worker to send the `{ event: "exit" }` message before destroying its proxy sockets.

* The worker waits for the parent to disconnect.
* The parent waits for the worker to send `"exit"`.

This results in an infinite loop that only breaks when the parent's 5-second graceful shutdown timeout forcefully terminates the worker thread.

### The Fix at nitropack

By calling `listener?.close()` asynchronously without wrapping it in an `await` Promise, the shutdown sequence proceeds and immediately sends the `"exit"` event to the parent. The parent receives the signal, successfully closes the proxy connections, and the listener naturally drains and closes in milliseconds. This completely eliminates the 5-second delay and prevents orphaned resources.

Changes at [nitro-dev.ts shutdown](https://github.com/nitrojs/nitro/blob/v2/src/presets/_nitro/runtime/nitro-dev.ts#L137-L144)
```ts
async function shutdown() {
  server.closeAllConnections?.();
  listener?.close();
  await nitroApp.hooks.callHook("close").catch(console.error);
  parentPort?.postMessage({ event: "exit" });
}
```

Applied these changes on my local:
```ts
async function shutdown() {
  console.log('🛑 [Worker] Shutdown started...');
  server.closeAllConnections?.();
  listener?.close();
  await nitroApp.hooks.callHook("close").then(() => console.log('🛑 [Worker] Nitro hooks called')).catch(console.error);
  console.log('🛑 [Worker] Sending "exit" event to the parent...');
  parentPort?.postMessage({ event: "exit" });
}
```

Nuxt cli quit correctly

userquin:
The changes in this PR won't be necessary if nitropack change shutdown logic:

nuxt cli quit command without changes in this pr but nitrpack patched

danielroe:
this seems like a nitro bug, rather than something that should be worked around here

cc: @pi0

PR #1539, opened 2026-09-18
fix(dev): show an app log once when a fork serves it

benjamincanac:
### 📚 Description

Once a fork is serving the app, every `console.log` from server code shows twice in the dev UI. The line reaches the parent by the log channel, and again because the worker's stdout goes through the fork's `consola.wrapAll()` and is forwarded as a second `nuxt:internal:dev:log`.

The UI already pairs a channel log with the same line caught on its way to the terminal, but only when that one is marked `raw`, and a fork's copy was not. The pairing also needed a `requestId`, so a log written outside a request could double in process too.

The fork now marks what its own consola caught as `raw`. In the event log the `paired` flag becomes the set of routes a log has arrived by: the app's report, a consola reporter, and the printed output. An entry takes one arrival from each in any order, so the same log is one entry and two logs saying the same thing stay two. A single flag was not enough here, with three arrivals some orders still left a duplicate.

The runtime wrapper still calls the original `console` method, so nothing changes without the dev UI. Output written straight to `process.stdout` from the worker still shows once, with and without `--fork`.

I checked this in `playground-nightly` under a pty in both modes. The e2e there does not catch it as is, since vitest leaks `NODE_ENV=test` into the CLI, which puts consola at level 1 and leaves `--fork` off.


danielroe:
is this still true with the latest nightly?

danielroe:
oh, maybe I fixed this in a branch I'm working on

danielroe (approved):

